PR #6, opened 2019-06-17
First approach to modularized layers

Artyom17:
I worked mostly on "multi-layer.md", however, I have some changes in layers-core.md and new README.md too. Thus, layers-core.md is not really ready for review yet, but feel free to provide me with any feedback (including the layers-core.md).


multi-layer.md
```diff
@@ -0,0 +1,130 @@
...
+* Layers are drawn in the same order as they are specified in via `XRSession/updateRenderState`, with the 0th layer drawn first. Layers are drawn with a "painter’s algorithm," with each successive layer potentially overwriting the destination layers whether or not the new layers are virtually closer to the viewer. There is no depth testing between the layers. Layer's visibility is controlled by presence (or absence) of the layer in the sequence. To hide or show a layer - call [updateRenderState](https://immersive-web.github.io/webxr/#dom-xrsession-updaterenderstate) and provide a new sequence of layers.
+
+* Add common properties to [XRLayer](https://immersive-web.github.io/webxr/#xrlayer-interface) (such as chromatic aberration or alpha blending flags):
+* * `chromaticAberrationCorrection` - boolean, enables optical chromatic aberration correction for the layer when not done by default; chromatic aberration correction improves visual fidelity of the layer's rendering, but it might be computationally expensive; this flag may be ignored by the browser, if this feature is not supported.
```
klausw: This seems rather low-level - on what basis should an application decide if chromatic aberration correction should be on or off? I think there's a fairly specific range of hardware performance where the platform has sufficient GPU resources to do chromatic aberration correction, but not enough to do so unconditionally for all layers. 

If you do want to keep this as a hint, please clarify that implementations are free to ignore the "correction off" value if it's not a performance concern, i.e. for a desktop VR system. As written, it could be interpreted as meaning that an implementation that supports chromatic aberration correction would be required to respect this flag, and that's likely not intended.

Bigger picture, is there a need for a more general performance hint system, i.e. providing optimization hints ("optimize this layer for best pixel quality") or feedback from the system on per-layer cost for certain feature sets? For example, a platform might support multiple layers, but performance would degrade if there's too many of them or if there's too much screen area overlap.

(This isn't intended to derail the discussion, but I think it's worth keeping an eye on the risk of apps tuning experiences based on user agent sniffing or similar due to lack of more direct feedback mechanisms.)

multi-layer.md
```diff
@@ -0,0 +1,130 @@
...
+    newRenderState.layers[1] = new XRLayerQuad(session, gl);
+  }
+}
+session.updateRenderState(newRenderState);
+
+```
+
+### Setting common layers properties
+```javascript
+  if (maxLayerCount > 1 && session.isLayerSupported("XRLayerQuad"))
```
klausw: I'm not sure about the boolean "isLayerSupported" - that's a hard question for the implementation to answer if it doesn't yet know what flags will be set on the layer and what size they'll be.

Alternatively, would it make sense to have a layer configuration attribute set that's provided to both the query function and the layer initializer? That way, there's less risk of inconsistency, and the implementation could potentially respond with "I could do this, but it's unlikely to hit target framerates". That could either be on a layer-by-layer basis, or potentially as a "here's my proposed overall layer configuration" all-at-once call.

README.md
```diff
@@ -29,97 +29,6 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
 Another example where composition layers are going to shine is displaying text or high resolution textures in the virtual world. Since composition layers allow to sample source texture at full resolution without downsampling to the resolution of the eye buffers (which is usually much lower than the physical resolution of the device's screen(s)) the readability of the text is going to be significantly improved.
```
klausw: >  usually much lower than the physical resolution of the device's screen(s)

FWIW, this phrasing seems a bit mobile-centric. On desktop VR systems, the eye buffers are in some cases higher resolution than the device screen. The overall point is of course still valid, and there's a quality hit from repeated resampling.

multi-layer.md
Artyom17: Sounds good, will try to make the change

README.md
Artyom17: Agreed. Will make the change to wording.

PR #13, opened 2020-02-11
Sketching out the F2F proposal in explainer form. 

toji:
Includes some tweaks made since then to address various issues I thought of in the proposal as presented.

Probably the biggest change from the F2F is the addition of the `XRGraphicsBindings` concept, which is used here primarily to allow the `XRLayerSource` creation to be asynchronous but my intent would be to also use it for things like retrieving the cube map for AR light estimation, since that was once of the sticking points for me in that proposal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,4 +28,294 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
 
 Another example where composition layers are going to shine is displaying text or high resolution textures in the virtual world. Since composition layers allow to sample source texture at full resolution without multiple re-samplings the readability of the text is going to be significantly improved.
 
+## Usage
 
+Using the Layers API consists of three primary steps:
+
+  1. Creating, positioning, and sizing the layers to be shown.
+  1. Creating graphics API resources to back the layers.
+  1. Rendering to the graphics API resources during the frame loop.
+
+### Layer creation
+
+Layers are created by the `XRSession`, using the `requestLayer()` series of fuctions, patterned after the `requestReferenceSpace()` method. Unlike other `request...` methods this function does not return a `Promise`, since it's expected that layer construction will be fairly lightweight. It may return `null`, however, if the `XRSession` does not support the requested layer type. The only layer type that _must_ be supported is `"projection"`. The returned instance depends on the requested type, with each layer type corresponding to it's own interfaces which extends `XRLayer`.
+
+```js
+// Returns an XRProjectionLayer instance
+const projectionLayer = xrSesssion.requestLayer('projection');
+```
+
+This explainer will not cover the details of every possible layer. For additional details please see the separate explainers for each layer type:
+
+  - XRProjectionLayer
+  - [XRQuadLayer](./quad.md)
+  - [XRCylinderLayer](./cylinder.md)
+  - [XREquirectLayer](./equirect.md)
+  - [XRCubeLayer](./cubemap.md)
+
+Layers are not presented to the XR device until they have been added to the `layers` `XRRenderState` property with the `updateRenderState()` method. Setting the `layers` array will override the `baseLayer` if one is present, with `baseLayer` reporting the first element in the `layers` array. Layers will be presented in the order they are given in the `layers` array, with each layers being given in "back-to-front" order. Layers may have alpha blending applied if the layer's `blendSourceAlpha` attribute is `true`, but no depth testing may be performed between layers.
+
+```js
+const projectionLayer = xrSesssion.requestLayer('projection');
+const quadLayer = xrSesssion.requestLayer('quad');
+
+xrSession.updateRenderState({ layers: [projectionLayer, quadLayer] });
+```
+
+`updateRenderState()` _may_ throw an exception if more layers are specified than the `XRSession` supports simultaneously.
+
+### Graphics API binding
+
+On their own, layers have no visual data, and will be ignored during compositing until provided with a backing `XRLayerSource`. A layer source is typically a GPU resource, like a texture, provided by one of the Web platform's graphics APIs. In order to specify which API is providing the layer sources an `XRGraphicsBinding` instance for the API in question must be created. For example, creating a graphics binding fow WebGL would function like this:
+
+```js
+const canvas = document.createElement('canvas');
+const gl = canvas.getContext('webgl', { xrCompatible: true });
+const xrGfx = new XRWebGLGraphicsBinding(xrSession, gl);
+```
+
+A session that wanted to take advantage of WebGL 2.0 techniques would alternatively create a `XRWebGL2GraphicsBinding` like so:
+
+```js
+const canvas = document.createElement('canvas');
+const gl = canvas.getContext('webgl2', { xrCompatible: true });
+const xrGfx = new XRWebGL2GraphicsBinding(xrSession, gl);
+```
+
+And a theoretical WebGPU graphics binding (which is not being proposed at this time and is offered only for illustrative purposes) may look like this:
+
+```js
+const gpuAdapter = await navigator.gpu.requestAdapter({ xrCompatible: true });
+const gpuDevice = await gpuAdapter.requestDevice();
+const xrGfx = new XRWebGPUGraphicsBinding(xrSession, gpuDevice);
+```
+
+Each graphics API may have unique requirements that must be satisfied before a context can be used in the creation of an `XRGraphicsBinding`. For example, a `WebGLRenderingContext` must have its `xrCompatible` flag set prior to being passed to the constructor of the `XRWebGLGraphicsBinding` instance.
+
+Any interaction between the `XRSession` the graphics API, such as allocating or retrieving textures, will go through this `XRGraphicsBinding` instance, and the exact mechanics of the interaction will typically be API specific. This allows the rest of the WebXR API to be graphics API agnostic and more easily adapt to future advances in rendering techniques.
+
+## Layer source allocation
+
+Once an `XRGraphicsBinding` instance has been acquired, it can be used to create backing `XRLayerSource`s for each layer. What form this takes depends on the API being used, but is generally expected to be a texture or other render target expressed using the APIs native interfaces.
+
+In all cases a `set____LayerSource` method will be called with the layer that the layer source is being created for, as well as any additional information needed to create the appropriate graphics resources. The method will return a promise that will resolve to the associated `XRLayerSource` type once the graphics resources have been created and the layer is ready to begin presenting images provided by the new layer source.
+
+```js
+const xrGfx = new XRWebGLGraphicsBinding(xrSession, gl);
+
+const layer = xrSession.requestLayer('projection');
+const source = await xrGfx.setFramebufferLayerSource(layer, { alpha: false });
+```
+
+If a layer source had previously been created for the passed layer it will become detached and the associated graphics resources discarded once the new layer source has been resolved. Creationg of new layer sources must not resolve until any pending `XRFrame` has been processed.
+
+Depending on the layer type that the layer source is being created for, additional data may need to be provided in order to allocate graphics resources appropriately. For instance, layer types other than `"projection"` must be given an explicit pixel width and height, as well as whether or not the image should be stereo or mono. This is because those properties cannot be inferred from the hardware or layer type as they can with a `"projection"` layer.
+
+```js
+const xrGfx = new XRWebGLGraphicsBinding(xrSession, gl);
+
+const layer = xrSession.requestLayer('quad');
+const isStereo = true;
+const source = await xrGfx.setFramebufferLayerSource(layer, 1024, 768, { stereo: true });
+```
+
+Passing `true` for stereo here indicates that you are able to provide stereo imagery for this layer source, but if the XR device is unable to display stereo imagery it may automatically force the layer to be created as mono instead to reduce memory and rendering overhead. Layers that are created as mono will never be automatically changed to stereo, regardless of hardware capabilities.
+
+## Rendering
+
+During `XRFrame` processing each layer source can be updated with new imagery. Calling `getViewSubImage()` with a view from the `XRFrame` will return an `XRSubImage` indicating what section of the associated graphics resources will be presented to the view's associated physical display. If a layer source has not yet been set for the layer `getSubImage()` with return an `XRSubImage` with all values set to zero or false.
+
+Every layer source will also have a `getCurrent_____()` method that retrieves the graphics resources that should be rendered to for the current `XRFrame`. Calling this method indicates that the imagery should be updated, otherwise any previous rendering will be displayed again, reprojected if necessary.
+
+```js
+// Render Loop for a projection layer with a WebGL framebuffer source.
+const xrGfx = new XRWebGLGraphicsBinding(xrSession, gl);
+
+const layer = xrSession.requestLayer('projection');
+const source = await xrGfx.setFramebufferLayerSource(layer);
+
+xrSession.updateRenderState({ layers: [layer] });
+xrSession.requestAnimationFrame(onXRFrame);
+
+function onXRFrame(time, xrFrame) {
+  xrSession.requestAnimationFrame(onXRFrame);
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,4 +28,294 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
...
+```js
+const projectionLayer = xrSesssion.requestLayer('projection');
+const quadLayer = xrSesssion.requestLayer('quad');
+
+xrSession.updateRenderState({ layers: [projectionLayer, quadLayer] });
+```
+
+`updateRenderState()` _may_ throw an exception if more layers are specified than the `XRSession` supports simultaneously.
+
+### Graphics API binding
```
cabanier: Is this really needed?

README.md
```diff
@@ -28,4 +28,294 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
...
+const gpuAdapter = await navigator.gpu.requestAdapter({ xrCompatible: true });
+const gpuDevice = await gpuAdapter.requestDevice();
+const xrGfx = new XRWebGPUGraphicsBinding(xrSession, gpuDevice);
+```
+
+Each graphics API may have unique requirements that must be satisfied before a context can be used in the creation of an `XRGraphicsBinding`. For example, a `WebGLRenderingContext` must have its `xrCompatible` flag set prior to being passed to the constructor of the `XRWebGLGraphicsBinding` instance.
+
+Any interaction between the `XRSession` the graphics API, such as allocating or retrieving textures, will go through this `XRGraphicsBinding` instance, and the exact mechanics of the interaction will typically be API specific. This allows the rest of the WebXR API to be graphics API agnostic and more easily adapt to future advances in rendering techniques.
+
+## Layer source allocation
```
cabanier: I still think that we should try to find a way to combine layer and graphics backing at the same time.
Otherwise we have to deal with all the edge case (ie what happens if a layer has no backing or gets backing later, what if multiple layer sources use the same layer)

README.md
```diff
@@ -28,4 +28,294 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
...
+  for (let view in xrViewerPose.views) {
+    let viewport = layer.getViewSubImage(view).viewport;
+    gl.viewport(viewport.x, viewport.y, viewport.width, viewport.height);
+    
+    // Render from the viewpoint of xrView
+  }
+}
+```
+
+In some cases, such as a mono `XRQuadLayer` being shown on a stereo device, multiple `XRView`s may return the same `XRSubImage` values. To avoid rendering the same view multiple times in these scenarios the `primary` attribute of the `XRSubImage` will be be set to `false` for all but one of the overlapping views. (It should be noted that `XRProjectionLayer`s will never return `XRSubImage`s with `primary` set to false.)
```
cabanier: I think mono layers should have a different method since they are typically not dependent on the view

README.md
```diff
@@ -28,4 +28,294 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
...
+  "quad",
+  "cylinder",
+  "equirect",
+  "cube"
+}
+
+interface XRLayer {
+  XRSubImage? getViewSubImage(XRView view);
+
+  XRReferenceSpace referenceSpace;
```
cabanier: projection layers don't need a reference space. Maybe move this down?

README.md
Artyom17: What is more interesting is the multiview case, where you render both views in one pass. We need a way to handle this.

README.md
Artyom17: What is more interesting is the multiview case, where you render both views in one pass. We need a way to handle this.

README.md
Artyom17: I am on opposite side, I really love separation of the graphics side from the layers themselves. Yes, the API becomes a bit more complex, but in long term it worth it, IMO.

README.md
Artyom17: It is a factory for layer sources and it makes sense (to me).

README.md
cabanier: you will be able to do this. There is nothing stopping you from looking at both subimages at the same time.

README.md
cabanier: Why not have it on session? 
On the web platform, decorator classes are typically not used.

README.md
cabanier: In C++ I would agree with you. 
However, this is javascript inside the browser. Usually interfaces are designed to *ONLY* do what is required for specific use cases. 
For instance, OpenXR allows any combination of layers, poses and swapchains regardless if they make sense. On the Web we should only allow constructs that are sensical.

README.md
Artyom17: I was a bit confused here. So, 'primary' is the way to check for mono, but it is not really working for mutliview. One thing when you render one view (mono), and another thing when you render multiview (stereo, both views, but in one pass).

I think we just need a sample for mutliview case here, something like:

```js
// Render Loop for a projection layer with a WebGL framebuffer source.
const xrGfx = new XRWebGL2GraphicsBinding(xrSession, gl);
const layer = xrSession.requestLayer('projection');
const source = await xrGfx.setTextureArrayLayerSource(layer);
const mv_ext = gl.getExtension('OVR_multiview2');
xrSession.updateRenderState({ layers: [layer] });
xrSession.requestAnimationFrame(onXRFrame);

let fbo = gl.createFramebuffer();

function onXRFrame(time, xrFrame) {
  xrSession.requestAnimationFrame(onXRFrame);
  gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
  if (mv_ext != null) {
    // multiview rendering
    let colorTexture = source.getCurrentColorTexture()
    mv_ext.framebufferTextureMultiviewOVR(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, colorTexture, 0, 0, 2);
    let leftView = xrViewerPose.views[0];
    let rightView = xrViewerPose.views[1];
    let leftProj = leftView.projectionMatrix;
    let rightProj = rightView.projectionMatrix;
    let viewport = layer.getViewSubImage(leftView).viewport; // viewports should be the same for left & right eyes
    gl.viewport(viewport.x, viewport.y, viewport.width, viewport.height);
    ....
  } else {
    // non-multiview rendering, using texture array
    for (let view in xrViewerPose.views) {
      let viewport = layer.getViewSubImage(view).viewport;
      gl.viewport(viewport.x, viewport.y, viewport.width, viewport.height);
    
      // Render from the viewpoint of xrView
    }
  }
}
```

README.md
```diff
@@ -28,4 +28,294 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
...
+  WebGLTexture? getCurrentDepthStencilTexture();
+}
+```
+
+## Appendix B: Open Questions
+
+  - Sensible defaults of all the layer values.
+  - Do we need an attribte to communicate max number of supported layers?
+  - Do we need to report on the layer source whether or not it's mono/stereo, since the XR device may force a downgrade?
+  - Depth: Should we consider allowing a separate layer for this like OpenXR does, or continue pattern of allowing it's use by default (implies most layer types need to be able to allocate both a color and depth/stencil buffer)
```
Artyom17: Depth wise - I think, we can include the depth into the layer source for simplicity, which existence can be controlled via 'depth' (or 'has_depth') attribute. By doing this, we would also be able to discard it at the end to avoid depth resolve.

Artyom17: I don't see anything about stereo layout here (side-by-side, top-to-bottom, or just stereo (for texture arrays)), but it could be added later.

Artyom17: Will need to add more rendering-hinting flags, such as MSAA samples, FFR, CAC.

Artyom17: Will need to have a way to signal that layer's image has changed or not: thus we might avoid extra GPU work and just reuse the image from the previous frame. In fact, since all of those 'textures' and 'framebuffers' are actually swapchains, this is a must to have, since the underlying compositor must now which slice of the swapchain to use and whether it needs to be incremented or not.

README.md
Artyom17: I don't know, actually, I think either way is fine for me as long as we keep the graphics API <-> layer separation. If we put it in session, then the session class becomes kinda bloated and the XRSession now must be aware of all of those Graphics APIs.... This is what @toji are trying to avoid, I assume.

README.md
Artyom17: One more thing: are all possible LayerSources should be supported? What is the way to determine which ones are and which ones are not?

cabanier: We need to design the API in such a way that non-sensical combinations can't be constructed.
Otherwise we have to define in prose which ones are valid and have to throw on invalid combinations.

cabanier: I can't comment your previous example. What would the view/projection matrices be for non-projection layers? Are there frameworks that can render stereo in equirect and cube layers?

toji:
Thanks for the feedback thus far! Been thinking about it this evening and think there's an approach that preserves some of the qualities I'm looking for while addressing some of the primary concerns I've heard thus far. Specifically reducing layer creation to a single call/interface and still enforce that some invalid layer combinations can't be created. Basically it turns the `XRGraphicsBinding` interfaces I proposed (a name which I'm not super fond of, BTW) into a layer factory, and has developers pass the layer into a `getCurrentFramebuffer(layer)` or `getCurrentTexture(layer)` method in the frame loop.

I didn't do anything to address feedback about stereo/mono and the kinda clunky `primary` sub image cruft yet. I agree they could be handled better, though.

cabanier:
I really like your latest changes. Let's merge them and then we can continue iterating

README.md
```diff
@@ -32,41 +32,13 @@ Another example where composition layers are going to shine is displaying text o
...
-
-xrSession.updateRenderState({ layers: [projectionLayer, quadLayer] });
-```
-
-`updateRenderState()` _may_ throw an exception if more layers are specified than the `XRSession` supports simultaneously.
-
 ### Graphics API binding
 
-On their own, layers have no visual data, and will be ignored during compositing until provided with a backing `XRLayerSource`. A layer source is typically a GPU resource, like a texture, provided by one of the Web platform's graphics APIs. In order to specify which API is providing the layer sources an `XRGraphicsBinding` instance for the API in question must be created. For example, creating a graphics binding fow WebGL would function like this:
+When a layers is created it is typically backed a GPU resource, like a texture, provided by one of the Web platform's graphics APIs. In order to specify which API is providing the layer's GPU resources an `XRGraphicsBinding` instance for the API in question must be created. For example, creating a graphics binding for WebGL would function like this:
```
Artyom17: Well, now I have some issues with this change. How do we specify which type of the resources (framebuffer, texture, texture array, etc) is being used for the layer?

README.md
```diff
@@ -28,4 +28,263 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
...
+quadLayer.width = 1.5;
+quadLayer.height = 1.5;
+```
+
+This explainer will not cover the details of positioning and shaping every possible layer. For additional details please see the separate explainers for each layer type:
+
+  - [XRQuadLayer](./quad.md)
+  - [XRCylinderLayer](./cylinder.md)
+  - [XREquirectLayer](./equirect.md)
+  - [XRCubeLayer](./cubemap.md)
```
Manishearth: Probably should have a document (even a placeholder) for projection layer too.

README.md
```diff
@@ -28,4 +28,263 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
...
+interface XRCubeLayer extends XRLayer {
+  XRReferenceSpace referenceSpace;
+  attribute DOMPoint orientation;
+}
+
+//
+// Graphics Bindings
+//
+
+dictionary XRWebGLLayerSourceInit {
```
Manishearth: Should this just be `: XRWebGLLayerInit`?

README.md
cabanier: We'd have to fix that next

README.md
```diff
@@ -94,53 +66,79 @@ Each graphics API may have unique requirements that must be satisfied before a c
...
+The various layer types are created with the  `request____Layer` series of methods on the `XRGraphicsBinding` instance. Information about the graphics resources required, such as whether or not to allocate a depth buffer or alpha channel, are passed in at layer creation time and will be immutable for the lifetime of the layer. The method will return a promise that will resolve to the associated `XRLayer` type once the graphics resources have been created and the layer is ready to be displayed.
 
 ```js
 const xrGfx = new XRWebGLGraphicsBinding(xrSession, gl);
+const layer = await xrGfx.requestProjectionLayer({ alpha: false });
+```
+
+Layer types other than an `XRProjectionLayer` must be given an explicit pixel width and height, as well as whether or not the image should be stereo or mono. This is because those properties cannot be inferred from the hardware or layer type as they can with an `XRProjectionLayer`.
```
Artyom17: Note, that the projection layer may be backed by different GPU resources. I'd love to see how that gonna work now w/o layer sources.

README.md
```diff
@@ -28,4 +28,263 @@ One of the very common use cases of WebXR is 180 and/or 360 photos or videos, bo
...
+
+//
+// Layer types
+//
+
+interface XRProjectionLayer extends XRLayer {
+}
+
+interface XRQuadLayer extends XRLayer {
+  attribute XRRigidTransform transform;
```
Manishearth: doesn't have to be resolved here but I suggested during the F2F that we just use offset spaces, reducing the number of places where we're doing matrix math. We might want to file a followup once this lands.

README.md
cabanier: What do you mean by "different"? A different gl context?

cabanier:
@toji you linked to my old Magic Leap github account. 
I will see if I can get it moved to my fb account

README.md
toji: In this case the resource type would be dictated by the graphics binding you created the layer with, and each graphics binding would support only a single type. ie: The `XRWebGLGraphicsBinding` would always back layers it creates with `WebGLFramebuffers`. (Which also happens to be why that binding doesn't have a method for creating cube layers, because I don't think you can handle that effectively with a single framebuffer.)

This is a change from what I had previously, where a WebGL2 binding could create both framebuffers or textures, but it seemed like a reasonable simplification. Note that there's nothing here that would prevent you from using two different graphics bindings simultaneously, so you could still mix and match.

Another change that this introduces is that in order to reshape the GPU resources you'd have to recreate the entire layer, but the layer source was always the expensive part of that anyway so I don't think that recreating the layer+source (effectively) here is actually much different. Just means the user would have to copy over the layer positioning attribs.

README.md
toji: No, because that's an existing dictionary with attribs that don't quite match what's needed here. Please note that I think this is a terrible name, by the way, and we can definitely do something cleaner, I just haven't figured out what yet.

README.md
```diff
@@ -94,53 +66,79 @@ Each graphics API may have unique requirements that must be satisfied before a c
...
+const layer = xrGfx.requestProjectionLayer();
 
 xrSession.updateRenderState({ layers: [layer] });
 xrSession.requestAnimationFrame(onXRFrame);
 
 function onXRFrame(time, xrFrame) {
   xrSession.requestAnimationFrame(onXRFrame);
 
-  gl.bindFramebuffer(gl.FRAMEBUFFER, source.getCurrentFramebuffer());
+  gl.bindFramebuffer(gl.FRAMEBUFFER, xrGfx.getCurrentFramebuffer(layer));
```
Artyom17: How is it known that there IS a current framebuffer? I don't see that it was requested anyhow....

toji: The fact that it has a framebuffer backing it is due to the fact that the `XRWebGLGraphicsBinding` is the thing that created the layer. I think that interface name is unnecessarily confusing things though. If it was called, for example, `XRWebGLFramebufferLayerFactory` that would probably help with comprehension of how this pattern is intended to work.

toji:
> @toji you linked to my old Magic Leap github account.
> I will see if I can get it moved to my fb account

Oops! Sorry! I've sent your other account an invite to the Immersive Web org. Let me know if you want me to remove the Magic Leap account's membership so that people don't accidentally send more requests to it.

README.md
Artyom17: Textures, framebuffers, texture arrays. But @toji already responded above.

PR #14, opened 2020-02-11
Moved GPU accessors to subImage

cabanier:
It feels a bit cleaner to move the WebGL buffers to subImage.
subImage tells you what texture and region of texture to use so why not put it all in 1 location

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,8 +253,8 @@ interface XRCubeLayer extends XRLayer {
 //
 
 dictionary XRLayerInit {
-  unsigned int pixelWidth;
-  unsigned int pixelHeight;
+  required unsigned int pixelWidth;
+  required unsigned int pixelHeight;
   boolean stereo = false;
   boolean depth = true;
   boolean stencil = false;
--- a/README.md
+++ b/README.md
@@ -191,17 +191,15 @@ function onXRFrame(time, xrFrame) {
 
 dictionary XRSubImage {
   XRViewport viewport;
-  unsigned long? imageIndex; // for texture array
-  boolean primary;
+  boolean primary; // ??
   WebGLTexture? colorTexture;
-  WebGLTexture? depthStencilTextur;
+  WebGLTexture? depthStencilTexture;
   WebGLFramebuffer? framebuffer;
 }
 
 interface XRLayer {
   readonly attribute unsigned long pixelWidth;
   readonly attribute unsigned long pixelHeight;
-  readonly attribute unsigned long arraySize;
   readonly attribute boolean ignoreDepthValues;
 
   boolean blendTextureSourceAlpha = false;
@@ -281,7 +279,7 @@ interface XRWebGL2GraphicsBinding {
   Promise<XREquirectLayer> requestEquirectLayer(XRLayerInit init);
   Promise<XRCubeLayer> requestCubeLayer(XRLayerInit init); // Note only available with WebGL 2
   
-  XRSubImage? getViewSubImage(XRLayer layer, XRView view);
+  XRSubImage? getViewSubImage(XRLayer layer);
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -138,9 +138,10 @@ xrSession.requestAnimationFrame(onXRFrame);
 function onXRFrame(time, xrFrame) {
   xrSession.requestAnimationFrame(onXRFrame);
 
-  gl.bindFramebuffer(gl.FRAMEBUFFER, xrGfx.getCurrentFramebuffer(layer));
   for (let view in xrViewerPose.views) {
-    let viewport = layer.getViewSubImage(view).viewport;
+    let subImage = layer.getViewSubImage(view)
+    gl.bindFramebuffer(gl.FRAMEBUFFER, subImage.frameBuffer);
+    let viewport = subImage.viewport;
     gl.viewport(viewport.x, viewport.y, viewport.width, viewport.height);
     
     // Render from the viewpoint of xrView
@@ -167,11 +168,12 @@ xrSession.requestAnimationFrame(onXRFrame);
 function onXRFrame(time, xrFrame) {
   xrSession.requestAnimationFrame(onXRFrame);
 
-  gl.bindFramebuffer(gl.FRAMEBUFFER, xrGfx.getCurrentFramebuffer(quadLayer));
   for (let view in xrViewerPose.views) {
     let subImage = quadLayer.getViewSubImage(view);
 
     if (subImage.primary) {
+      gl.bindFramebuffer(gl.FRAMEBUFFER, subImage.framebuffer);
+      let viewport = subImage.viewport;
       gl.viewport(viewport.x, viewport.y, viewport.width, viewport.height);
       
       // Render content for view.eye
@@ -189,8 +191,11 @@ function onXRFrame(time, xrFrame) {
 
 dictionary XRSubImage {
   XRViewport viewport;
-  unsigned long imageIndex;
+  unsigned long? imageIndex; // for texture array
   boolean primary;
+  WebGLTexture? colorTexture;
+  WebGLTexture? depthStencilTextur;
+  WebGLFramebuffer? framebuffer;
 }
 
 interface XRLayer {
@@ -248,6 +253,7 @@ interface XRCubeLayer extends XRLayer {
 dictionary XRLayerInit {
   required unsigned int pixelWidth;
   required unsigned int pixelHeight;
+  boolean framebuffer = true; // should it be false?
   boolean stereo = false;
   boolean depth = true;
   boolean stencil = false;
@@ -264,8 +270,6 @@ interface XRWebGLGraphicsBinding {
   Promise<XRQuadLayer> requestQuadLayer(XRLayerInit init);
   Promise<XRCylinderLayer> requestCylinderLayer(XRLayerInit init);
   Promise<XREquirectLayer> requestEquirectLayer(XRLayerInit init);
-
-  WebGLFramebuffer? getCurrentFramebuffer(XRLayer layer);
 }
 
 interface XRWebGL2GraphicsBinding {
@@ -276,9 +280,6 @@ interface XRWebGL2GraphicsBinding {
   Promise<XRCylinderLayer> requestCylinderLayer(XRLayerInit init);
   Promise<XREquirectLayer> requestEquirectLayer(XRLayerInit init);
   Promise<XRCubeLayer> requestCubeLayer(XRLayerInit init); // Note only available with WebGL 2
-
-  WebGLTexture? getCurrentColorTexture(XRLayer layer);
-  WebGLTexture? getCurrentDepthStencilTexture(XRLayer layer);
 }
 ```
 
@@ -287,4 +288,5 @@ interface XRWebGL2GraphicsBinding {
   - Sensible defaults of all the layer values.
   - Do we need an attribte to communicate max number of supported layers?
   - Do we need to report on the layer source whether or not it's mono/stereo, since the XR device may force a downgrade?
-  - Depth: Should we consider allowing a separate layer for this like OpenXR does, or continue pattern of allowing it's use by default (implies most layer types need to be able to allocate both a color and depth/stencil buffer)
+  - Depth: Should we consider allowing a separate layer for this like OpenXR does, or continue pattern of allowing it's use by default (implies most layer types need to be able to allocate both a color and depth/stencil buffer)-
+  - Should there be a way to explicitly dispose of a layer?
--- a/README.md
+++ b/README.md
@@ -253,8 +253,8 @@ interface XRCubeLayer extends XRLayer {
 //
 
 dictionary XRLayerInit {
-  required unsigned int pixelWidth;
-  required unsigned int pixelHeight;
+  unsigned int pixelWidth;
+  unsigned int pixelHeight;
   boolean stereo = false;
   boolean depth = true;
   boolean stencil = false;
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ function onXRFrame(time, xrFrame) {
   xrSession.requestAnimationFrame(onXRFrame);
 
   for (let view in xrViewerPose.views) {
-    let subImage = layer.getViewSubImage(view)
+    let subImage = xrGfx.getViewSubImage(layer, view)
     gl.bindFramebuffer(gl.FRAMEBUFFER, subImage.frameBuffer);
     let viewport = subImage.viewport;
     gl.viewport(viewport.x, viewport.y, viewport.width, viewport.height);
@@ -169,7 +169,7 @@ function onXRFrame(time, xrFrame) {
   xrSession.requestAnimationFrame(onXRFrame);
 
   for (let view in xrViewerPose.views) {
-    let subImage = quadLayer.getViewSubImage(view);
+    let subImage = xrGfx.getViewSubImage(quadLayer, view);
 
     if (subImage.primary) {
       gl.bindFramebuffer(gl.FRAMEBUFFER, subImage.framebuffer);
@@ -199,8 +199,6 @@ dictionary XRSubImage {
 }
 
 interface XRLayer {
-  XRSubImage? getViewSubImage(XRView view);
-
   readonly attribute unsigned long pixelWidth;
   readonly attribute unsigned long pixelHeight;
   readonly attribute unsigned long arraySize;
@@ -270,6 +268,8 @@ interface XRWebGLGraphicsBinding {
   Promise<XRQuadLayer> requestQuadLayer(XRLayerInit init);
   Promise<XRCylinderLayer> requestCylinderLayer(XRLayerInit init);
   Promise<XREquirectLayer> requestEquirectLayer(XRLayerInit init);
+  
+  XRSubImage? getViewSubImage(XRLayer layer, XRView view);
 }
 
 interface XRWebGL2GraphicsBinding {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -187,6 +189,15 @@ function onXRFrame(time, xrFrame) {
 // Layer interface
 //
 
+dictionary XRSubImage {
```
cabanier: I changed it to a dictionary since it's just a holder of information. 
If possible we should avoid adding classes to the global namespace

README.md
```diff
@@ -245,9 +250,10 @@ interface XRCubeLayer extends XRLayer {
 // Graphics Bindings
 //
 
-dictionary XRWebGLLayerSourceInit {
+dictionary XRLayerInit {
```
cabanier: Shortened the name

README.md
```diff
@@ -287,4 +288,5 @@ interface XRWebGL2GraphicsBinding {
   - Sensible defaults of all the layer values.
   - Do we need an attribte to communicate max number of supported layers?
   - Do we need to report on the layer source whether or not it's mono/stereo, since the XR device may force a downgrade?
-  - Depth: Should we consider allowing a separate layer for this like OpenXR does, or continue pattern of allowing it's use by default (implies most layer types need to be able to allocate both a color and depth/stencil buffer)
+  - Depth: Should we consider allowing a separate layer for this like OpenXR does, or continue pattern of allowing it's use by default (implies most layer types need to be able to allocate both a color and depth/stencil buffer)-
+  - Should there be a way to explicitly dispose of a layer?
```
cabanier: Layer would be eventually cleaned up by the JS VM but that might take too long. 
I'm also unsure if JS knows how much memory they take up.

toji: That sounds like a really good idea to me.

README.md
toji: I tend to agree with you, but I've heard from platform ergonomics folks/tag that the general pattern they expect from the web is dictionaries for inputs, classes for outputs. 🤷‍♀

README.md
toji: I'm 100% fine with this as long as we do a bit of validation to make sure it applies across all the layer types we want to support except projection, which feels like it's "special". Cube map is the only one I really question. We also want to make sure it applies to WebGPU as well (I'm pretty sure it does.)

toji:
I see what you're going for, but I'm a bit skeptical that this pattern is better for developers. The primary thing that I question is that it looks like every view/sub image could have a completely different texture/framebuffer, which would both increase render loop overhead (you now have to move the framebuffer binding inside the view loop) but also makes it much more difficult to confidently do OVR_multiview rendering. We could spec it, of course, to say that each sub image exposes the same underlying GPU resource, but in that case it does feel like we should be exposing that at a higher level of the interface hierarchy.

README.md
cabanier: Looking at some specs (ie https://www.w3.org/TR/webrtc/), there are functions that return dictionaries.
It's a bit weird to introduce classes just to satisfy some platform convention. I'd really like to know why the tag wants this as it's not very javascript-y.

toji: I'll see if I can get some additional feedback on it.

cabanier:
> The primary thing that I question is that it looks like every view/sub image could have a completely different texture/framebuffer, which would both increase render loop overhead (you now have to move the framebuffer binding inside the view loop) 

That is correct. On WebGl 1, we want to return an array of textures; one for each eye.
Only WebXR style side-by-side would return the same texture. Would that extra binding call create any overhead? Since the texture was already bound it seems like it's a no-op

> but also makes it much more difficult to confidently do OVR_multiview rendering. We could spec it, of course, to say that each sub image exposes the same underlying GPU resource, but in that case it does feel like we should be exposing that at a higher level of the interface hierarchy.

WebGL 2 would always return the same texture for a layer. This would negate the need to call getViewSubImage per eye.

I think this is aligning well with the XRWebGLGraphicsBinding/XRWebGL2GraphicsBinding split you made

README.md
Manishearth: I can't find any part in WebRTC that does this? But yes, this is a platform convention.

Manishearth: Furthermore, webidl isn't really _equipped_ to deal with this, I suspect browser codegen might even choke on this.

Remember that null dictionaries are equivalent to dictionaries with all empty fields, and we're returning a nullable in `getViewSubImage()`, and that's probably not going to work the way we want it to.

I think webidl parsers will choke on the `WebGLTexture?` below as well.

README.md
Manishearth: Should we make this inherit from `XRWebGLLayerInit`?

Or perhaps create a new superclass dictionary type that everyone can inherit from? Perhaps that can be called `XRLayerInit` instead?

Manishearth:
I don't have strong opinions on the actual change being made here.

README.md
cabanier: > I can't find any part in WebRTC that does this? But yes, this is a platform convention.

See the getParameters call [here](https://www.w3.org/TR/webrtc/#rtcrtpsender-interface)

cabanier: > I think webidl parsers will choke on the `WebGLTexture?` below as well.

I keep getting confused about webidl. :-)

I just want to indicate that these are optional parameters. How do we do that? If we can't, we can just return them as null.

README.md
cabanier: > Should we make this inherit from `XRWebGLLayerInit`?

We're missing a couple of field from `XRWebGLLayerInit`: `antialias` and `framebufferScaleFactor`

> perhaps create a new superclass dictionary type that everyone can inherit from? Perhaps that can be called XRLayerInit instead?

Do you want `XRWebGLLayerInit` inherit from this base class as well?

README.md
Manishearth: Perhaps `optional WebGLTexture`. I'm not sure, like I said dictionaries aren't really designed to be returned, and the framework around them is optimized for the fields being supplied, not consumed.

README.md
Manishearth: Yeah, basically. We can make this change in the core spec and inherit from it here IMO. Doesn't have to be done now.

But it does mean that perhaps XRLayerInit should be reserved for the superclass.

cabanier:
> I don't have strong opinions on the actual change being made here.

@toji do you feel more comfortable with my change to move the subimage upstream?

toji:
Sorry, but I still don't feel comfortable that this is the right pattern. A few more points of concern:

For one, I looked back through some previous notes and, combined with Manish's feedback, do feel like we should stick to the platform's preferred convention of returning interfaces not dictionaries. I know some APIs don't follow this (WebGL is another example) but it was made pretty clear to me in the past that those outliers were not seen favorably.

In terms of data structure, I think having every possible GPU resource type on the `XRSubImage` when only one of them is valid at a time may cause confusion and makes the interface more bloated. A possible alternative is to have one return type for each GPU resource type. That's a little opaque from the point of view of the IDL, but at least it avoids conflating different API's resources and it would allow us to customize the returned values for the resource type in question (Framebuffers don't need an array index, for example.)

```webidl
interface XRSubImage {
  readonly attribute XRViewport viewport;
}

interface XRWebGLFramebufferSubImage {
  readonly attribute WebGLFramebuffer framebuffer;
}

interface XRWebGLTextureSubImage {
  readonly attribute unsigned long imageIndex;
  readonly attribute WebGLTexture colorTexture;
  readonly attribute WebGLTexture depthTexture;
}

// Etc.

partial interface XRProjectionLayer {
  XRSubImage? getViewSubImage(view);
}
```

However, beyond that I've been looking through your previous feedback on my PR/email and considering what you said about mono vs. stereo layers and I agree that the design I proposed regarding using sub images to drive rendering of mono content may be overly complicated. It's possible that we can drop `getViewSubImage()` from non-projection layers entirely, which would be a nice simplification, but also would mean that we'd need to get the GPU resources from somewhere else.

cabanier:
> For one, I looked back through some previous notes and, combined with Manish's feedback, do feel like we should stick to the platform's preferred convention of returning interfaces not dictionaries. I know some APIs don't follow this (WebGL is another example) but it was made pretty clear to me in the past that those outliers were not seen favorably.

OK. I reverted that change.

> In terms of data structure, I think having every possible GPU resource type on the `XRSubImage` when only one of them is valid at a time may cause confusion and makes the interface more bloated. A possible alternative is to have one return type for each GPU resource type. That's a little opaque from the point of view of the IDL, but at least it avoids conflating different API's resources and it would allow us to customize the returned values for the resource type in question (Framebuffers don't need an array index, for example.)

I agree. That does look cleaner. I added that to my change.

> However, beyond that I've been looking through your previous feedback on my PR/email and considering what you said about mono vs. stereo layers and I agree that the design I proposed regarding using sub images to drive rendering of mono content may be overly complicated. It's possible that we can drop `getViewSubImage()` from non-projection layers entirely, which would be a nice simplification, but also would mean that we'd need to get the GPU resources from somewhere else.

I agree. This is why I added it to XRWebGLGraphicsBinding/XRWebGL2GraphicsBinding since it knows what flavor of GPU resources to produce.
For mono or texture array layers, we don't care about what eye we're rendering so I removed `XRView` from the function call.

README.md
```diff
@@ -261,24 +269,25 @@ interface XRWebGLGraphicsBinding {
...
-  Promise<XRCylinderLayer> requestCylinderLayer(XRWebGLLayerSourceInit init);
-  Promise<XREquirectLayer> requestEquirectLayer(XRWebGLLayerSourceInit init);
-
-  WebGLFramebuffer? getCurrentFramebuffer(XRLayer layer);
+  Promise<XRQuadLayer> requestQuadLayer(XRLayerInit init);
+  Promise<XRCylinderLayer> requestCylinderLayer(XRLayerInit init);
+  Promise<XREquirectLayer> requestEquirectLayer(XRLayerInit init);
+  
+  XRSubImage? getViewSubImage(XRLayer layer); // for mono layers
+  XRSubImage? getViewSubImage(XRLayer layer, XRView view); // for stereo layers
```
toji: Oh! I missed earlier that you moved these methods off of the layer types and onto here. If that's the case, we can have them return the appropriate `XRSubImage` type directly, which removes some of the opaqueness. 👍

README.md
```diff
@@ -245,9 +252,10 @@ interface XRCubeLayer extends XRLayer {
 // Graphics Bindings
 //
 
-dictionary XRWebGLLayerSourceInit {
+dictionary XRLayerInit {
   required unsigned int pixelWidth;
   required unsigned int pixelHeight;
+  boolean framebuffer = true; // should it be false?
```
toji: I'm not sure what this is providing? Whether or not you get a framebuffer or texture would be determined by which GraphicsBinding type you're constructing the layers with. As I mentioned to Artem in the previous PR, they would probably be less confusing if they were called something like `XRWebGLFramebufferLayerFactory` and `XRWebGL2TextureLayerFactory` respectively.

cabanier: Does that mean we'll have 4 graphics binding objects? Or could be get by with only 2?(Texture/Framebuffer and switch on what gl flavor is passed in)

toji: Only 2. WebGL 1 will only deal with Framebuffers, because it's the only way you can get any antialiasing. WebGL 2 can do both, but if it wants to do Framebuffers instead of textures it should just use the WebGL 1 bindings.

(Okay, I lied: We'll have 3 once WebGPU arrives. It will only do textures.)

cabanier: I see. I was under the impression we needed textures for webgl 1. Maybe @Artyom17  can chime in on what we need.
If it's not needed, it can indeed be simplified.

Artyom17: @toji is correct that framebuffer is the only way to render with AA, however, afaik, it is also true for WebGL 2 (there is no direct render-to-texture-multisampling way (yet? we are discussing the corresponding WebGL extension based on EXT_multisampled_render_to_texture GL ES extension at the moment; and that extension could work in WebGL 1 too).
Otherwise, framebufferTexture2D can be still used with WebGL 1, as well as texImage2D (for videos or images). So, I'd say we need textures support for WebGL 1.

cabanier: ok. So maybe we can get by with `XRWebGLFramebufferLayerFactory` and `XRWebGLTextureLayerFactory` but have different behavior depending on the gl context.

For WebGL1, you would get an array of textures and for WebGL2, it would be a texture array

README.md
```diff
@@ -245,9 +250,10 @@ interface XRCubeLayer extends XRLayer {
 // Graphics Bindings
 //
 
-dictionary XRWebGLLayerSourceInit {
+dictionary XRLayerInit {
   required unsigned int pixelWidth;
```
Artyom17: I am not sure pixelWidth/Height should be marked as 'required' (unless I misunderstand what this actually means). For example, for projection layer it is not required. But scalefactor might be.

cabanier: fixed.
I think scaleFactor should likely move to `getViewSubImage`

README.md
toji: @Artyom17 Hm... I didn't realize that you still wanted the ability to use textures with WebGL 1. That complicates things a bit.

Now I'm curious about something: Do you see antialiasing as a necessity for Quad/Cylinder/Equirect layers?

README.md
toji: That would be too late. The scaleFactor only applies to projection layers, and is used when allocating the swapchain, so it has to be supplied at layer creation time. It means "Scale the recommended projection layer size by this much".

toji: Oh, and the pixelWidth/Height IS required for non-projection layers, which in the current IDL is the only place that dictionary is used. Projection layers were using the old XRWebGLLayerInit dict.

cabanier: I was under the impression that this particular scalefactor would tell you by how much to upscale your rendered area.
For instance, you might want to allocate your cylinder layer quite large in case it is close to the user but when the user walks away, you want to optimize the experience and maybe render at half the resolution.

In that case you'd pass scalefactor of .5 and get back a smaller viewport. Then you render to a smaller area and when the swap happens, the compositor will know by how much to upscale.

This would only be useful for cylinder/quad layers I suspect.

cabanier: > Oh, and the pixelWidth/Height IS required for non-projection layers, which in the current IDL is the only place that dictionary is used. Projection layers were using the old XRWebGLLayerInit dict.

Yes! I will change it back

README.md
Artyom17: @toji so, you are asking if we need a framebuffer support for quad/cyl/equirect? You mean, overall, for either WebGL 1 or 2? I probably can live without framebuffer support for Quad/Cyl/Equirects, hoping that AA WebGL R-2-T extensions are going to be accepted....

cabanier: @Artyom17 is it acceptable that Projection and Cube layers are *only* constructed with framebuffer backends and Quad, Cylinder, Equirect with texture backends?

cabanier:
@toji do you have fundamental objections to the latest changes? If not, I will update the samples to the latest idl and ask @Artyom17 to merge.

README.md
toji: I don't think Cube layers can be represented by a framebuffer. At least not a single one. The way that you render to them is to bind each face as the framebuffer color buffer separately.

toji:
Give me a moment to review. I didn't understand Oculus' needs for access to framebuffer/texture surfaces until last night, and I want to think through this again with that in mind.

README.md
cabanier: I could be done with a side-by-side framebuffer similar to the other layers. I don't know if this is done in practice.
Maybe *only* projection layers have framebuffer backend.

cabanier:
> Give me a moment to review. I didn't understand Oculus' needs for access to framebuffer/texture surfaces until last night, and I want to think through this again with that in mind.

ok! There were some good comments in this thread that I want to split up as issues

README.md
```diff
@@ -187,24 +189,29 @@ function onXRFrame(time, xrFrame) {
...
+}
+
+interface XRWebGLFramebufferSubImage : XRSubImage {
+  readonly attribute WebGLFramebuffer framebuffer;
+}
 
+interface XRWebGLTextureSubImage : XRSubImage {
+  readonly attribute unsigned long imageIndex;
+  readonly attribute WebGLTexture colorTexture;
+  readonly attribute WebGLTexture depthTexture;
```
toji: This should be nullable and the name should more accurately be `depthStencilTexture`, since if a stencil channel is requested it will have to be exposed here.

README.md
```diff
@@ -255,30 +262,31 @@ dictionary XRWebGLLayerSourceInit {
...
-  Promise<XRCylinderLayer> requestCylinderLayer(XRWebGLLayerSourceInit init);
-  Promise<XREquirectLayer> requestEquirectLayer(XRWebGLLayerSourceInit init);
-  Promise<XRCubeLayer> requestCubeLayer(XRWebGLLayerSourceInit init); // Note only available with WebGL 2
-
-  WebGLTexture? getCurrentColorTexture(XRLayer layer);
-  WebGLTexture? getCurrentDepthStencilTexture(XRLayer layer);
+  Promise<XRQuadLayer> requestQuadLayer(XRLayerInit init);
+  Promise<XRCylinderLayer> requestCylinderLayer(XRLayerInit init);
+  Promise<XREquirectLayer> requestEquirectLayer(XRLayerInit init);
+  Promise<XRCubeLayer> requestCubeLayer(XRLayerInit init); // Note only available with WebGL 2
```
toji: This needs to move to the `XRWebGLTextureLayerFactory`, since you can't represent a cube map with a single framebuffer, but can with a single texture.

README.md
toji: I've been wondering about that myself, which is part of why I asked about antialiasing. It's worth starting up a separate conversation about.

PR #18, opened 2020-02-13
Removed the concept of the framebuffer layer factory

### LINKED ISSUE #17: Are WebGLFramebuffers needed for layer types other than projection?
Something of a continuation of [a comment thread](https://github.com/immersive-web/layers/pull/14#discussion_r379035243) from #14. For the sake of simplification, I'd like to examine whether or not it's possible for us to only use the current opaque framebuffer surfaces for projection layers, and force everything else to use textures in their respective APIs.

The primary benefit of doing so would be allowing the layer creation implementation and use to be significantly simpler and reduce the amount of layer factories needed.

The primary downside would be that any non-projection layers in WebGL 1.0 would not have antialiasing, and in WebGL 2.0 they would need to rely on blitting from multisampled framebuffers for antialiasing until multisample render to texture extensions are available. (WebGPU will have multisampled textures as a core feature.)

If this was an acceptable compromise, the way I see it working is like this: Either we'd drop the `XRWebGLFramebufferLayerFactory` entirely, and rely solely on the current `XRWebGLLayer` interface to provide framebuffer-backed projection layers OR we'd add a special-case framebuffer projection layer creation function to the `XRWebGLTextureLayerFactory` (which could probably be renamed to something simpler because we no longer have to differentiate.) Then the remaining factory would take a GLenum for the texture type desired in the Projection, Quad, Cylinder, and Equirect factory methods. This would allow the developer to choose between gl.TEXTURE_2D or (if they created the factory with a WebGL 2.0 context) gl.TEXTURE_2D_ARRAY as the backing texture type. (Cube layers would always return gl.TEXTURE_CUBE_MAP textures) The fact that the user specifies it means that there's less confusion about the internal format and what framebuffer binding methods they need to use.

Possible IDL:

```webidl
interface XRWebGLLayerFactory {
  constructor(XRSession session, XRWebGLRenderingContext context);

  double getNativeProjectionScaleFactor();

  Promise<XRProjectionLayer> requestProjectionLayer(GLenum textureTarget, XRProjectionLayerInit init); // Note different dictionary
  Promise<XRQuadLayer> requestQuadLayer(GLenum textureTarget, XRLayerInit init);
  Promise<XRCylinderLayer> requestCylinderLayer(GLenum textureTarget, XRLayerInit init);
  Promise<XREquirectLayer> requestEquirectLayer(GLenum textureTarget, XRLayerInit init);
  Promise<XRCubeLayer> requestCubeLayer(XRLayerInit init);

  XRWebGLTextureSubImage? getViewSubImage(XRLayer layer); // for mono layers
  XRWebGLTextureSubImage? getViewSubImage(XRLayer layer, XRView view); // for stereo layers
}
```

The `XRLayer` types don't need to change to accommodate this, but we do get to drop the `XRWebGLFramebufferSubImage` interface. We'd also maybe want a new `XRProjectionLayerInit` dictionary because the existing `XRWebGLLayerInit` dictionary explicitly mentions framebuffer scale, which would no longer be an accurate term, and antialiasing, which is not applicable to the texture layers.

Whether or not this works for us is largely based on what kind of content we expect developers to put into non-projection layers and how willing we believe them to be to use WebGL 2.0 when antialiasing is a must. In my experience quad and cylinder layers are frequently populated with pre-rendered images or videos, for which hardware multisampling won't matter. (Equirect/Cube layers seem to be almost exclusively used with prerendered content.) UIs seem to be the other primary use, and so that would hit some limitations, but they may be largely mitigated by finding a way to enable DOM content in layers, at which point the browser's renderer is expected to do any antialiasing for us. The remaining cases can treat antialiasing of those layers as a quality enhancement that can be provided by using WebGL 2.0.

Thoughts?

toji:
/fixes #17

Simplifies the layers by leaving use of opaque framebuffers exclusively to the `XRWebGLLayer`, which is still treated as a first-class layer type here and as the only way to get antialiasing in WebGL 1.0. All new layer types use `WebGLTexture`s, which may be requested as `TEXTURE_2D` or `TEXTURE_2D_ARRAY` as the developer needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,9 @@ For some non-projection layers, such as a mono `XRQuadLayer` being shown on a st
 ```js
 // Render Loop for a projection layer with a WebGL framebuffer source.
 const glLayerFactory = new XRWebGLLayerFactory(xrSession, gl);
-const quadLayer = xrGfx.requestQuadLayer(gl.TEXTURE_2D, { pixelWidth: 512, pixelHeight: 512, stereo: false });
+const quadLayer = glLayerFactory.requestQuadLayer(gl.TEXTURE_2D, {
+  pixelWidth: 512, pixelHeight: 512, stereo: false
+});
 // Position 2 meters away from the origin with a width and height of 1.5 meters
 quadLayer.referenceSpace = xrReferenceSpace;
 quadLayer.transform = new XRRigidTransform({z: -2});
@@ -215,7 +217,7 @@ function onXRFrame(time, xrFrame) {
   xrSession.requestAnimationFrame(onXRFrame);
 
   gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
-  let subImage = xrGfx.getSubImage(quadLayer);
+  let subImage = glLayerFactory.getSubImage(quadLayer);
   gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0,
     subImage.colorTexture, 0);
   let viewport = subImage.viewport;
--- a/README.md
+++ b/README.md
@@ -38,62 +38,68 @@ Using the Layers API consists of three primary steps:
 
 ### Graphics API binding
 
-When a layers is created it is typically backed a GPU resource, like a texture, provided by one of the Web platform's graphics APIs. In order to specify which API is providing the layer's GPU resources an `XRGraphicsBinding` instance for the API in question must be created. For example, creating a graphics binding for WebGL would function like this:
+When a layer is created it is backed by a GPU resource, typically a texture, provided by one of the Web platform's graphics APIs. In order to specify which API is providing the layer's GPU resources an Layer Factory for the API in question must be created. For example, creating a layer factory for WebGL would function like this:
 
 ```js
 const canvas = document.createElement('canvas');
 const gl = canvas.getContext('webgl', { xrCompatible: true });
-const xrGfx = new XRWebGLFramebufferLayerFactory(xrSession, gl);
+const glLayerFactory = new XRWebGLLayerFactory(xrSession, gl);
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -123,60 +131,97 @@ xrSession.updateRenderState({ layers: [projectionLayer, quadLayer] });
...
 
-  for (let view in xrViewerPose.views) {
-    let subImage = xrGfx.getViewSubImage(quadLayer, view);
-
-    if (subImage.primary) {
-      gl.bindFramebuffer(gl.FRAMEBUFFER, subImage.framebuffer);
-      let viewport = subImage.viewport;
-      gl.viewport(viewport.x, viewport.y, viewport.width, viewport.height);
+  gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
+  let subImage = xrGfx.getSubImage(quadLayer);
```
Artyom17: xrGfx -> glLayerFactory

README.md
```diff
@@ -123,60 +131,97 @@ xrSession.updateRenderState({ layers: [projectionLayer, quadLayer] });
...
+}
+```
 
-const quadLayer = xrGfx.requestQuadLayer({ pixelWidth: 512, pixelHeight: 512, stereo: false });
+For some non-projection layers, such as a mono `XRQuadLayer` being shown on a stereo device, multiple `XRView`s may return the same `XRSubImage` values. To avoid rendering the same view multiple times in these scenarios mono layers can be rendered just once using the `XRSubImage` provided by the the layer factory's `getSubImage()` method.
+
+```js
+// Render Loop for a projection layer with a WebGL framebuffer source.
+const glLayerFactory = new XRWebGLLayerFactory(xrSession, gl);
+const quadLayer = xrGfx.requestQuadLayer(gl.TEXTURE_2D, { pixelWidth: 512, pixelHeight: 512, stereo: false });
```
Artyom17: xrGfx -> glLayerFactory

toji:
Done.

PR #26, opened 2020-03-03
Fix WebGL1 example

cabanier:
Referenced in https://github.com/immersive-web/webxr/issues/968

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ xrSession.updateRenderState({ layers: [projectionLayer, quadLayer] });
 
 During `XRFrame` processing each layer can be updated with new imagery. Calling `getViewSubImage()` with a view from the `XRFrame` will return an `XRSubImage` indicating the textures to use as the render target and what portion of the texture will be presented to the `XRView`'s associated physical display.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -133,7 +133,7 @@ xrSession.updateRenderState({ layers: [projectionLayer, quadLayer] });
 
 During `XRFrame` processing each layer can be updated with new imagery. Calling `getViewSubImage()` with a view from the `XRFrame` will return an `XRSubImage` indicating the textures to use as the render target and what portion of the texture will be presented to the `XRView`'s associated physical display.
 
-WebGL layers allocated with the `TEXTURE_2D` target will provide sub images with a unique, non-overlapping `viewport` and `imageIndex` of `0` for each `XRView`.
+WebGL layers allocated with the `TEXTURE_2D` target will provide sub images with a `viewport` and an `imageIndex` of `0` for each `XRView`.
```
toji: Let's put a note here that the texture may be unique for each sub image, and another one in the TEXTURE_2D_ARRAY section indicating that the texture will be the same for each sub image (though still with a different `imageIndex`)

PR #29, opened 2020-03-06
A proposal for video layers.

### LINKED ISSUE #22: Provide native support for video layers
Equirect and possible cylinder and quad layers should be contructable with a video element for efficient decoding of video streams.
What is the best API surface for this?

toji:
/fixes #22

Putting this up as a pull request to allow for easy commenting, but it's really just a proposal that I'd like to discuss.

The root idea here is pretty straightforward: Just have another layer factory that's specifically for video. (I've left the interface name a little more generic as "media" so that we could potentially use it for image tags later, but it's obviously not the focus right now.) There's a few variants of this approach that could work out well, though.

For one, given that we don't need any additional information besides the `XRSession` you could make an argument that these layer constructors belong directly on the `XRSession` itself. I'm not sure how I feel about that. I have a knee-jerk reaction to adding what feels like a lot of clutter to the session, and it may give the impression that videos are a "favored" media type for WebXR, but that's probably overthinking it. I do like the symmetry it offers with the WebGL/WebGPU layers. Also, there might be some UA benefits to be had from the need to explicitly ask for the media layer factory, since it could allow you to spin up certain decoder pathways/resources at that time, but I'm really not sure if that actually matters.

In terms of the initialization parameters, I'm not sure how simple we can get, especially when it comes to the layout. The most general approach I know of is still specifying a rect for the left and right eye, and overlapping rects are used for mono. That may be overkill, though, if ~all videos are a 50/50 split left-right or top-bottom. I've got an `invertStereo` boolean in there two in the case that most videos have the, for example, top = left eye and bottom = right eye, but again if the conventions are really strongly baked into the existing media pipelines then maybe that's not necessary. The more concerning thing for me is more experimental stuff like Google's 180 stereo format or the pyramid layout Facebook tried a while back, neither of which was trivially laid out in rectangular regions. (Although, to my knowledge neither of those formats actually gained much traction either.)

We can always tell people to fall back to WebGL rendering for exotic layouts, though, and if any non-rectangular layouts DO become a widespread standard in the future we can add them to the `XRMediaLayout` enum. This should be reasonably backwards compatible because attempting to use a new enum with an older implementation will throw an error at layer creation time, which is easy enough to try/catch and fall back to a WebGL path for.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,13 +148,13 @@ function onXRFrame(time, xrFrame) {
   xrSession.requestAnimationFrame(onXRFrame);
 
   gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
- 
+
   for (let view in xrViewerPose.views) {
     let subImage = glLayerFactory.getViewSubImage(layer, view);
     gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0,
       gl.TEXTURE_2D, subImage.colorTexture, 0);
     gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.DEPTH_ATTACHMENT,
-      gl.TEXTURE_2D, subImage.depthStencilTexture, 0);  
+      gl.TEXTURE_2D, subImage.depthStencilTexture, 0);
     let viewport = subImage.viewport;
     gl.viewport(viewport.x, viewport.y, viewport.width, viewport.height);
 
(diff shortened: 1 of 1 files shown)


cabanier:
This looks awesome! Unless people object, I will merge this tomorrow.
I agree with your feeling about potentially having these methods directly on the session. We can see if people feel strongly about this. 

I can see what the status is of the "pyramid layout" at Oculus.

PR #34, opened 2020-03-27
First pass at an editor's draft for WebXR layers

cabanier:
I started a skeleton spec. 
Even though it's not complete, I think it's ready for comments in case it's wildly wrong.
I will continue to make updates.

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -2,7 +2,7 @@
 Shortname: webxrlayers-1
 Title: WebXR Layers API Level 1
 Group: immersivewebwg
-Status: DREAM
+Status: ED
 TR:
 ED:
 Previous Version:
@@ -21,6 +21,11 @@ spec:infra;
 </pre>
 
 <pre class="anchors">
+spec: WebGL; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.0/
+    type: interface; text: WebGLFramebuffer; url: WebGLFramebuffer
+    type: interface; text: WebGLTexture; url: WebGLTexture
+    type: interface; text: WebGLRenderingContext; url: WebGLRenderingContext
+    type: interface; text: WebGLRenderingContextBase; url: WebGLRenderingContextBase
 </pre>
 
 <link rel="icon" type="image/png" sizes="32x32" href="favicon-32x32.png">
@@ -105,12 +110,9 @@ The spec adds support of `composition` layers to the WebXR spec. The benefits of
  * <b>Latency</b> Pose sampling for composition layers may occur at the very end of the frame and then certain reprojection techniques could be used to update the layer's pose to
  match it with the most recent HMD pose. This may significantly reduce the effective latency for the layers' rendering and as a result improve overall experience.
 
-ISSUE: how are the controllers drawn? Should they be optional, always there or drawn by the author?
-
 ISSUE: how to do hit testing? Should it be done by the UA?
 
 </section>
-
 
 
 Terminology {#terminology}
@@ -190,6 +192,12 @@ The <dfn attribute for="XRLayer">chromaticAberrationCorrection</dfn> attribute e
 aberration correction for the layer when not done by default.
 
 {{destroy()}} will delete the underlying attachments. If there are no attachments, this function does nothing.
+
+Each {{XRLayer}} has a <dfn for="XRLayer">context</dfn> object which is an instance 
+of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.
+
+Each {{XRLayer}} has an associated <dfn for="XRLayer">session</dfn>, which is the 
+{{XRSession}} it was created with.
 
 XRProjectionLayer {#xrprojectionlayer}
 -----------------
@@ -308,7 +316,8 @@ A XREquirectLayer renders a layer where the [=XR Compositor=] renders directly f
 
 ISSUE: this section needs clarification
 
-<pre class="idl">interface XRCubeLayer : XRLayer {
+<pre class="idl">
+interface XRCubeLayer : XRLayer {
   readonly attribute boolean stereo;
   attribute XRReferenceSpace referenceSpace;
   attribute DOMPoint orientation;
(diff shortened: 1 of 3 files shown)


cabanier:
/agenda What is the process for putting the editor's draft in the repo?
Technically, I believe the editor needs to be approved by the group.

Manishearth:
[Rendered](https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/immersive-web/layers/ac16b436c509edaa77066f61e94ea81a85ababa9/webxrlayers-1.bs)

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+The spec adds support of `composition` layers to the WebXR spec. The benefits of layers are as follows:
+
+ * <b>Performance and judder</b> Composition layers are presented at the frame rate of the compositor (i.e. native refresh rate of HMD) rather than at the application frame rate.
+    Even when the application is not updating the layer's rendering at the native refresh rate of the compositor, the compositor might be able to re-project the existing
+    rendering to the proper pose. This results in smoother rendering and less judder. Another feature of layers is that each of them can have different resolution. This allows the
+    application to scale down the main eye buffer resolution on low performance systems, but keeping essential information, such as text or a map, in its own layer at a higher
+    resolution.
+
+ * <b>Legibility/vsisual fidelity</b> The resolution for eye-buffers for 3D world rendering can be set to relatively low values especially on low performance systems. It would
```
Manishearth: typo 
```suggestion
 * <b>Legibility/visual fidelity</b> The resolution for eye-buffers for 3D world rendering can be set to relatively low values especially on low performance systems. It would
```

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+Abstract: This specification describes support for accessing the geometry of real world objects during a WebXR session.
+</pre>
+
+<pre class="link-defaults">
+spec:infra;
+    type:dfn; text:string
+</pre>
+
+<pre class="anchors">
```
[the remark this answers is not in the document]
cabanier: I based this on the WebXR spec.
Should it be changed there as well?

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+<pre highlight="js">
+navigator.xr.requestSession('immersive-vr', {
+    optionalFeatures: ['layers']
+  }
+</pre>
+</div>
+
+Layers are only supported for XRSessions created with XRSessionMode of {{XRSessionMode/"immersive-vr"}} 
+or {{XRSessionMode/"immersive-ar"}}. {{XRSessionMode/"inline"}} sessions must not support layers.
```
toji:
```suggestion
or {{XRSessionMode/"immersive-ar"}}. {{XRSessionMode/"inline"}} sessions MUST not support layers.
```

If this is intended to be a normative requirement (which it sounds like it is) then the all-caps 'MUST' should be used to indicate that.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+Layer types {#xrlayertypes}
+===========
+
+Mono and stereo layers {#monovsstereo}
+----------------------
+A stereo layer must supply a different view to render to for each eye. Stereo layers should be refreshed
```
toji: > Stereo layers should be refreshed close to the device's native frame rate.

I'm not sure if this is the message we want to communicate? Seems obvious you're talking about projection layers here, which do offer stereo content, but when I think of "stereo" layers I also consider stereo video or even stereo still images to be part of that. Obviously those are unlikely to need to be refreshed at the headset's rate. I wonder if it's worth listing projection as a distinct type from stereo and mono, noting that it may contain stereo imagery and should refresh near the native framerate, while the other two may refresh less often?

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+Layer types {#xrlayertypes}
+===========
+
+Mono and stereo layers {#monovsstereo}
+----------------------
+A stereo layer must supply a different view to render to for each eye. Stereo layers should be refreshed 
+close to the device's native frame rate.
+
+A mono layers must supply a single view which is shown to each eye. There is no requirement to refresh mono 
+layers repeatedly.
```
toji: Brief examples of what kind of content these may encompass would be useful.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+  void destroy();
+};
+</pre>
+
+The <dfn attribute for="XRLayer">pixelWidth</dfn> and <dfn attribute for="XRLayer">pixelHeight</dfn> attributes 
+return the width and height of the GL attachments, respectively. If there are no attachments, they will return 0.
+
+The <dfn attribute for="XRLayer">blendTextureSourceAlpha</dfn> attribute enables the layer’s texture alpha channel.
+
+The <dfn attribute for="XRLayer">chromaticAberrationCorrection</dfn> attribute enables optical chromatic
```
toji: We should somehow indicate that this is more of a hint and that the compositor may not be able to respect it, since various systems may either force it on or not support it at all.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+  attribute boolean blendTextureSourceAlpha;
+  attribute boolean chromaticAberrationCorrection;
+
+  void destroy();
+};
+</pre>
+
+The <dfn attribute for="XRLayer">pixelWidth</dfn> and <dfn attribute for="XRLayer">pixelHeight</dfn> attributes 
+return the width and height of the GL attachments, respectively. If there are no attachments, they will return 0.
```
toji: > If there are no attachments

What circumstances would we expect that to happen?

toji: Oh, I guess if the layer has been destroyed?

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+Each {{XRLayer}} has a <dfn for="XRLayer">context</dfn> object which is an instance 
+of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.
+
+Each {{XRLayer}} has an associated <dfn for="XRLayer">session</dfn>, which is the 
+{{XRSession}} it was created with.
+
+XRProjectionLayer {#xrprojectionlayertype}
+-----------------
+A XRProjectionLayer is a layer that fills the entire view of the observer.
```
toji:
```suggestion
An {{XRProjectionLayer}} is a layer that fills the entire view of the observer.
```

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+Projection layer must always run stereo and should be refreshed close to the device's native frame rate.
```
toji: > Projection layer must always run stereo

That's not accurate for phone AR or similar form factors. Would be more correct to say they must have a separate sub image for each XRView or something similar.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+<pre class="idl">
+interface XRProjectionLayer : XRLayer {
+  readonly attribute boolean ignoreDepthValues;
+};
+</pre>
+
+The <dfn attribute for="XRProjectionLayer">ignoreDepthValues</dfn> attribute, if <code>true</code>, indicates the 
+[=XR Compositor=] MUST NOT make use of values in the depth buffer attachment when rendering. When the attribute 
+is <code>false</code> it indicates that the content of the depth buffer attachment will be used by the
```
toji:
```suggestion
is <code>false</code> it indicates that the content of the depth buffer attachment can be used by the 
```

Minor change, but we shouldn't imply that all compositors will make use of the depth buffer.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+[=XR Compositor=] MUST NOT make use of values in the depth buffer attachment when rendering. When the attribute 
+is <code>false</code> it indicates that the content of the depth buffer attachment will be used by the 
+[=XR Compositor=] and is expected to be representative of the scene rendered into the layer.
+
+XRQuadLayer {#xrquadlayertype}
+-----------
+A XRQuadLayer renders a layer that takes up a flat rectangular space in the virtual environment.
+Only the front of the layer must be visible; the back face must not be drawn by the [=XR Compositor=].
+
+A XRQuadLayer has no thicknes. It is a two-dimensional object positioned and oriented in 3D space. The position
```
toji:
```suggestion
A XRQuadLayer has no thickness. It is a two-dimensional object positioned and oriented in 3D space. The position 
```

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+The <dfn attribute for="XRQuadLayer">width</dfn> and <dfn attribute for="XRQuadLayer">height</dfn> attributes 
+set and return the width and height of the layer in meters.
+
+XRCylinderLayer {#xrcylinderayertype}
+---------------
+A XRCylinderLayer renders a layer that takes up a curved rectangular space in the virtual environment.
+Only the front of the layer must be visible; the back face must not be drawn by the [=XR Compositor=].
+
+A XRCylinderLayer has no thicknes. It is a two-dimensional object positioned and oriented in 3D space. The position
```
toji:
```suggestion
An {{XRCylinderLayer}} has no thickness. It is a two-dimensional object positioned and oriented in 3D space. The position 
```

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+The <dfn attribute for="XRCylinderLayer">centralAngle</dfn> attribute controls the angle in degrees of the visible section of the cylinder.
+It grows symmetrically around the 0 angle.
+
+The <dfn attribute for="XRCylinderLayer">aspectRatio</dfn> attribute controls the ratio of the visible cylinder section.
+
+<img src="images/cylinder_layer_params.png" style="width:80%"/>
+
+XREquirectLayer {#xrequirectlayertype}
+---------------
+A XREquirectLayer renders a layer where the [=XR Compositor=] must map an equirectangular coded data onto the inside of a sphere.
```
toji:
```suggestion
An {{XREquirectLayer}} renders a layer where the [=XR Compositor=] must map an equirectangular coded data onto the inside of a sphere.
```
I'm gonna stop flagging these for now, but in general we want to make sure any interface names like this are properly linked.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+The <dfn attribute for="XREquirectLayer">scaleX</dfn> and <dfn attribute for="XREquirectLayer">scaleY</dfn> attributes 
+set and return a scale of the texture coordinates after the mapping to 2D.
+
+The <dfn attribute for="XREquirectLayer">biasX</dfn> and <dfn attribute for="XREquirectLayer">biasY</dfn> attributes 
+set and return a bias of the texture coordinates after the mapping to 2D.
+
+XRCubeLayer {#xcubelayertype}
+-----------
+A XREquirectLayer renders a layer where the [=XR Compositor=] renders directly from a cubemap.
```
toji:
```suggestion
An {{XRCubeLayer}} renders a layer where the [=XR Compositor=] renders directly from a cubemap.
```

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+};
+</pre>
+
+The <dfn attribute for="XRWebGLSubImage">colorTexture</dfn> attribute returns the color {{WebGLTexture}} for the {{XRLayer}}.
+
+The <dfn attribute for="XRWebGLSubImage">depthStencilTexture</dfn> attribute returns the depth/stencil {{WebGLTexture}} for the {{XRLayer}}. 
+If the layer was created without depth/stencil, this attribute returns null.
+
+The <dfn attribute for="XRWebGLSubImage">imageIndex</dfn> attribute returns the offset into the texture array. Valid only for layers 
+that were requested with TEXTURE_2D_ARRAY.
```
toji: It's a bit of a pain to do, because the WebGL specs are not well formatted for it, but we should try to link WebGL enums and the like back to the specs where they are defined. See the WebXR spec's XRWebGLLayer section for examples.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+
+<div class="algorithm" data-algorithm="construct-webgl-layer">
+
+The <dfn constructor for="XRWebGLLayerFactory">XRWebGLLayerFactory(|session|, |context|)</dfn> constructor 
+MUST perform the following steps when invoked:
+
+  1. Let |factory| be a new {{XRWebGLLayerFactory}}
+  1. If |session|'s [=ended=] value is <code>true</code>, throw an {{InvalidStateError}} and abort these steps.
+  1. If |context| is lost, throw an {{InvalidStateError}} and abort these steps.
+  1. If |session| is not an [=immersive session=] or |context|'s [=XR compatible=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
```
toji: Just for consistency it seems like this line should be two separate steps, because the checks aren't really related to each other.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+  1. Initialize |factory|'s [=XRWebGLLayer/context=] to |context|.
+  1. Initialize |factory|'s [=XRWebGLLayer/session=] to |session|.
+  1. Return |factory|.
+
+</div>
+
+ISSUE: can a user construct more than 1 factory?
```
toji: Yes, IMO, but we can discuss that elsewhere.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+};
+</pre>
+
+The <dfn dict-member for="XRProjectionLayerInit">depth</dfn> attribute defines if the {{XRProjectionLayer}} will have a depth buffer.
+
+The <dfn dict-member for="XRProjectionLayerInit">stencil</dfn> attribute defines if the {{XRProjectionLayer}} will have a stencil buffer.
+
+The <dfn dict-member for="XRProjectionLayerInit">alpha</dfn> attribute defines if the {{XRProjectionLayer}} will have an alpha channel.
+
+The <dfn dict-member for="XRProjectionLayerInit">scaleFactor</dfn> attribute defines the value that the |session|'s
```
toji: This description is accurate for the nativeScaleFactor, as you have below, but this `scaleFactor` is only the value that the recommended resolution should be multiplied by when allocating attachments. May want to change it to something along the lines of:

> The <dfn dict-member for="XRProjectionLayerInit">scaleFactor</dfn> attribute defines the value that the |session|'s [=recommended WebGL framebuffer resolution=] must be multiplied by when determining the resolution of the layer's attachments.

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+The {{nativeProjectionScaleFactor}} function returns the value that the |session|'s [=recommended WebGL framebuffer resolution=] 
+must be multiplied by to yield the |session|'s [=native WebGL framebuffer resolution=].
+
+<div class="algorithm" data-algorithm="requestProjectionLayerAlgo">
+The <dfn method for="XRWebGLLayerFactory">requestProjectionLayer(GLenum textureTarget, optional XRProjectionLayerInit init)</dfn> method creates a new {{XRProjectionLayer}}.
+
+When this method is invoked, it MUST run the following steps:
+
+  1. Let |promise| be [=a new Promise=].
+  1. If |session|'s [=ended=] value is <code>true</code>, throw an {{InvalidStateError}} and abort these steps.
```
toji: For any of the failure cases in an algorithm that returns a promise, the logic should follow this pattern. Changes needs to be made throughout this algorithm and the others.

```suggestion
  1. If |session|'s [=ended=] value is <code>true</code>, [=reject=] |promise| with an {{InvalidStateError}} and return it.
```

webxrlayers-1.bs
```diff
@@ -0,0 +1,730 @@
...
+  1. If |context| is lost, throw an {{InvalidStateError}} and abort these steps.
```
toji: Variables like |session| and |context| should be defined before being used, or if they're only used once should be explicit about which object's variable they are referring to. If this case it should be something like:

```
1. Let |session| be the target {{XRWebGLLayerFactory}}'s [=XRWebGLLayerFactory/session=].
1. Let |context| be the target {{XRWebGLLayerFactory}}'s [=XRWebGLLayerFactory/context=].

File: webxrlayers-1.bs [hunk 21]
```diff
@@ -0,0 +1,730 @@
...
+    1. Allocate and initialize resources compatible with |session|'s [=XRSession/XR device=], including GPU accessible memory buffers, as required to support the compositing of |layer|.
+    1. If |layer|’s resources were unable to be created for any reason, throw an {{OperationError}} and abort these steps.
+    1. [=/Resolve=] |promise| with |layer|.
+  1. Return |promise|.
+
+</div>
+
+ISSUE: define how scaleFactor works
+
+ISSUE: define "opaque" {{WebGLTexture}}
```
toji: We don't have any concept of opaque textures here, do we?

webxrlayers-1.bs
cabanier: done

webxrlayers-1.bs
cabanier: I plan on adding examples shortly

webxrlayers-1.bs
cabanier: done. Also I will add an issue as this attribute needs a better description.

webxrlayers-1.bs
cabanier: done

webxrlayers-1.bs
cabanier: I just removed this. 
Since all layers can now be now be mono or stereo, should we move the stereo attribute to XRLayer?

webxrlayers-1.bs
cabanier: I filed an issue on this earlier today.
I was thinking that the depthstenciltexure should only be provided on platforms where the depth buffer is used by the compositior.

webxrlayers-1.bs
cabanier: done

webxrlayers-1.bs
cabanier: done

webxrlayers-1.bs
cabanier: done

webxrlayers-1.bs
cabanier: done

webxrlayers-1.bs
cabanier: done

webxrlayers-1.bs
cabanier: no, not yet.

webxrlayers-1.bs
cabanier: done

webxrlayers-1.bs
cabanier: You 're right and I call out the refresh rate already for projection layers. 
Maybe I should remove that line?

AdaRoseCannon:
It seems that the bit didn't auto add the label, could you add it?

I've already sent out the email for this weeks call, please can you add the agenda label by hand and I'll make sure it gets added to next weeks call.

cabanier:
/agenda agree on WebXR layers draft & editors

Issue #35, opened 2020-03-30
Should the depth stencil buffer be optional


cabanier:
I updated the layers proposal so on platforms that don't use the depth buffer for reprojection, the depth/stencil texture is always null.
Was that the right thing to do?
/agenda

toji:
Hm... so the idea is that even if you ask for a depth buffer but it's not used in compositing it will return null? Not sure I'm on board with that, as it makes for more work for the developer when it doesn't seem like a large burden for the UA? Also may be a pit of failure where the developers hardware makes use of it and so they don't realize that they'll need to allocate their own textures in some cases.

I do wonder if there's a variant of this that would make more sense in this case, though? Since we're no longer doing the opaque framebuffer thing here we could say that if you DO request a depth buffer then you're expected to always ensure the values in it are appropriate for compositing. If you're not willing to do that then you can allocate your own depth buffers instead. Has the benefit of eliminating one of the more confusing options in the current `XRWebGLLayer`.

I will say that I'm a little concerned that devs who choose to allocate their own depth buffers won't have all the information needed to do so (number of views, specifically) until rAF callback has already started, which isn't the optimal point to allocate buffers. Maybe we want an event on the layer to indicate when the attachment shape (width, height, image count) changes?

cabanier:
> Hm... so the idea is that even if you ask for a depth buffer but it's not used in compositing it will return null?

Correct. That is what the spec currently says.

> Not sure I'm on board with that, as it makes for more work for the developer when it doesn't seem like a large burden for the UA? Also may be a pit of failure where the developers hardware makes use of it and so they don't realize that they'll need to allocate their own textures in some cases.

That is true which is why I raised this issue. If the hardware supports it and the author only tested on hardware that doesn't use depth for compositing, it will result in a sub-par experience.
On the other hand, it does allow the author to allocate their own depth buffer instead of relying on the default one that the UA would allocate.

> I do wonder if there's a variant of this that would make more sense in this case, though? Since we're no longer doing the opaque framebuffer thing here we could say that if you DO request a depth buffer then you're expected to always ensure the values in it are appropriate for compositing. If you're not willing to do that then you can allocate your own depth buffers instead. Has the benefit of eliminating one of the more confusing options in the current `XRWebGLLayer`.

Yes. I *think* this is another argument for leaving the spec as is.

[...]

PR #40, opened 2020-04-01
Depthstencil is now created by the UA if possible

cabanier:
Fix for #35


webxrlayers-1.bs
```diff
@@ -553,9 +555,10 @@ When this method is invoked, it MUST run the following steps:
          <dd> [=reject=] |promise| with an {{InvalidStateError}} and return it.
          <dd> 
          </dl>
+    1. let |depthsupport| be true if |context| is a {{WebGL2RenderingContext}} or the {{WEBGL_depth_texture}} extension is enabled in |context|.
```
toji, line 558: Ooh! This is a good check, thanks for considering it!

I think that down below if the user requests depth and `depthsupport` is `false` we should actually reject since we can't adequately fulfill the request. UAs will probably want to accompany it with a console error explaining why, but that's not part is not really something that we can/should spec.

PR #45, opened 2020-04-02
Added support for scalefactor

cabanier:


cabanier:
hmm, I have the green "Confirm merge" button. Will we lose something if I press it?

toji:
Huh... that's interesting. I'd say give it a go! We can look at the commit history to see if it does anything we didn't want.

toji:
I don't see anything unexpected being removed/changed in the [merge commit](https://github.com/immersive-web/layers/commit/a5cae6e221f3e3dc43c6786a83438ecbeaa125a4), so I think we're good!

cabanier:
and the spec was rebuilt as well.
Thanks for setting that up!

PR #47, opened 2020-04-04
Added notes to clarify XRWebGLLayerFactory and XRMediaLayerFactory behavior

cabanier:


webxrlayers-1.bs
```diff
@@ -500,6 +500,10 @@ of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.
 Each {{XRWebGLLayerFactory}} has an associated <dfn for="XRWebGLLayerFactory">session</dfn>, which is the 
 {{XRSession}} it was created with.
 
+NOTE: it is possible to create more than one {{XRWebGLLayerFactory}}. Any layer created with an instance of {{XRWebGLLayerFactory}} can be used
+ with layers created with other instances of {{XRWebGLLayerFactory}} as long as they were created with the same [=XRWebGLLayer/context=] and
```
toji, line 504: > as long as they were created with the same `context`

This somewhat implies that we wouldn't be able to mix WebGL layers and video layers, which I'm quite certain isn't what any of us actually want. What's the intent behind this restriction? I think it just needs to be worded differently.

cabanier: yes, my bad. `context` needs to be removed.

cabanier: @toji I ended up rewording both sections.
Let me know what you think.

webxrlayers-1.bs
```diff
@@ -726,8 +728,11 @@ should be inverted.
 
 XRMediaLayerFactory {#xrmedialayerfactorytype}
 -------------------
-The {{XRWebGLLayerFactory}} object is used to create layers that display the content of an {{HTMLVideoElement}}
-.
+The {{XRWebGLLayerFactory}} object is used to create layers that display the content of an {{HTMLVideoElement}}.
```
RafaelCintron, line 731: Shouldn't this sentence say refer to `XRMediaLayerFactory` instead of `XRWebGLLayerFactory`?  

`XRMediaLayerFactory` methods do not accept any WebGL things so I was surprised to see WebGL mentioned in the context of it.

webxrlayers-1.bs
```diff
@@ -726,8 +728,11 @@ should be inverted.
...
+
+NOTE: it is possible to create more than one {{XRMediaLayerFactory}}. The lifetime of layers is not tied to the lifetime of
```
RafaelCintron, line 733: Nit: The word "it" should be capitalized.

webxrlayers-1.bs
```diff
@@ -500,6 +500,10 @@ of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.
 Each {{XRWebGLLayerFactory}} has an associated <dfn for="XRWebGLLayerFactory">session</dfn>, which is the 
 {{XRSession}} it was created with.
 
+NOTE: it is possible to create more than one {{XRWebGLLayerFactory}}. Any layer created with an instance of {{XRWebGLLayerFactory}} can 
+be used with another instance of {{XRWebGLLayerFactory}} as long as both were created with the same [=XRWebGLLayer/session=]. The lifetime
```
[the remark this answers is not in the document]
cabanier: I had that restriction in the spec previously.
However, the way the spec is formulated, there's no need to the `XRWebGLRenderingContext` to be the same between the factories.

I do agree that that's a bit confusing so maybe I should put it back in.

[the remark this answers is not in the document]
RafaelCintron: I agree there is no need for `XRWebGLRenderingContext` to be the same between factories.  You just can't mix and resources between factories created with different rendering contexts or sessions.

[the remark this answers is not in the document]
cabanier: I don't feel strongly that only the session should be the same.
I'll update the note and the spec for layer creation to clarify.

cabanier:
Since this is holding up some refactoring, I will merge this PR. 
Please let me know if we need to make additional changes

Issue #39, opened 2020-04-01
Why does requestQuadLayer and requestProjection layer require a GL target?


RafaelCintron:
Examples from the explainer:
````JS
const layer = await glLayerFactory.requestProjectionLayer(gl.TEXTURE_2D_ARRAY, { alpha: false });
const layer = await glLayerFactory.requestQuadLayer(gl.TEXTURE_2D, { pixelWidth: 1024, pixelHeight: 768, stereo: true });
````

Why do `requestProjectionLayer` and `requestQuadLayer` require a GL target?  Are we intending to expand these to be more than `TEXTURE_2D` and `TEXTURE_2D_ARRAY` in the future?  

If the intent is to communicate whether the developer wants a single texture vs an array of textures, seems straightforward to add this as an item to the dictionary passed as the second parameter: `texturearray: true`.

As an aside, how many XR runtimes prefer side-by-side rendering these days? If there aren't any, my preference would be to have new APIs like these always return texture arrays.

cabanier:
> Why do `requestProjectionLayer` and `requestQuadLayer` require a GL target? Are we intending to expand these to be more than `TEXTURE_2D` and `TEXTURE_2D_ARRAY` in the future?
> If the intent is to communicate whether the developer wants a single texture vs an array of textures, seems straightforward to add this as an item to the dictionary passed as the second parameter: `texturearray: true`.

I don't think there are plans for any other objects. I'm happy to switch it to a boolean value.

> As an aside, how many XR runtimes prefer side-by-side rendering these days? If there aren't any, my preference would be to have new APIs like these always return texture arrays.

I *believe* we still do have a preference for side-by-side. @Artyom17 , what do you think?

cabanier:
/agenda should we switch to a book and/or drop support for side-by-side textures?

Manishearth:
I have a slight preference for making it into a boolean if we're never going to add more

Artyom17:
I don't like the idea of boolean, it is not informative and poorly readable. I'd vote for using an enum, like "texture", "texture-array".

RafaelCintron:
My strong preference is we have no enum or boolean and simply have layers standardize on texture arrays.  If we want to explore "cave" setups with more than two array entries, we can add an additional attribute where developers specify the number of layers on devices that support it.

Artyom17:
Uhm.... Isn't this a bit extreme? From one point of view I agree with you, this will simplify implementation of UA and promote WebGL 2 into WebXR.
[...]

PR #50, opened 2020-04-06
proposal to add XRLayerLayout

cabanier:
This enum allows control of the pixel layout for side-by-side textures


cabanier:
> I like this change overall! It does seem like there's some potential conflict with the discussion happening in #39 right now, though?

I submitted PR #51 as a solution for issue #39

> One thing that I'd be interested in figuring out both for this change and for the discussion in #39 is if it's a common scenario for media developers want to display to come pre-packaged as top/bottom or left/right stereo. I can see that if that was the case it would make it easier to do a single blit (though we would have to introduce limits on the viewports to make that case feasible). If not, though, I don't think there's too many benefits to the developer to specifying it this way.

Yes, that's why I added it to the agenda. I want to make sure everyone is ok with a more complex API to reduce the number of draw calls.

> Also wondering if we shouldn't enforce that "stereo" is separate textures at that point, to remove any ambiguity and allow developers to optimize their binding patterns.

For Oculus, we'd prefer to have side-by-side textures if possible since it reduces the number of calls and allocated resources.

> Finally, it feels an awful lot like we ought to be able to merge this and the `XRMediaLayout`. Maybe just specify that videos can't use the "stereo" enum?

That sounds reasonable. I'll make that change

cabanier:
> * In the explainer we previously said that for mono displays we would could force a layer requested as stereo down to mono. Does that still apply to stereo-left-right and stereo-top-bottom? Or should we always respect the layout and provide texture allocations that may be ignored by the compositor for the sake of allowing more consistent rendering paths? If the latter we probably want to let the developer know if the display is mono or not so they can manually allocate the right type if they're resource conscious.

Yes, it should be the latter. It's probably useful in general to know if a display is mono or stereo.
Maybe it could be added to WebXR?

> * "stereo-left-right" and "stereo-top-bottom" are nonsensical in cube maps, right? ("stereo" still applies, you'll just show a different cubemap to each eye.) We'll want to make sure we reject on that, same way we do video layers with "stereo".

Yes. I will make sure to address that. I filed issue #52  so we don't forget.

> * If the user does request "stereo" and the system decides to return that as a single texture, should we report the XRLayer's `layout` as "stereo-left-right" for the sake of accuracy? Could allow some minor binding pattern benefits.

That sounds very reasonable. @Artyom17 , what do you think?

Artyom17:
>> If the user does request "stereo" and the system decides to return that as a single texture, should we report the XRLayer's layout as "stereo-left-right" for the sake of accuracy? Could allow some minor binding pattern benefits.

Yes, I agree, it should report the actual layout.

>>"stereo-left-right" and "stereo-top-bottom" are nonsensical in cube maps, right? ("stereo" still applies, you'll just show a different cubemap to each eye.) We'll want to make sure we reject on that, same way we do video layers with "stereo".

Yes, this is correct. AFAIK, you can't choose texture regions on cube map.

>>Or should we always respect the layout and provide texture allocations that may be ignored by the compositor for the sake of allowing more consistent rendering paths? 

Layer creation should fail if the requested layout can't be provided. Otherwise, it will defeat the whole purpose to use the pre-rendered content.

cabanier:
@Artyom17 what should happen if you request a side-by-side texture with a texture array?
The spec currently throws an error.

Artyom17:
I think it is reasonable to either ignore (use it as 'stereo') or throw an error.

cabanier:
> I think it is reasonable to either ignore (use it as 'stereo') or throw an error.

ok. Since we should fail if the requested layout can't be provided, I'll leave the spec as-is.

PR #51, opened 2020-04-06
Added XRTextureType and updated algorithm + clean up of subimage creation

cabanier:
- Fix for #39. Got rid of WebGL value and replaced with enum
- broke duplicate validation algorithm out to simplify subimage creation steps

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -41,7 +41,7 @@ spec: WebGL; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.0/
     type: typedef; text: TEXTURE_2D; url: 5.14
 spec: WebGL 2.0; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/2.0/
     type: interface; text: WebGL2RenderingContext; url: WebGL2RenderingContext
-    type: typedef; text: TEXTURE_2D_ARRY; url: 3.7
+    type: typedef; text: TEXTURE_2D_ARRAY; url: 3.7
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -238,6 +238,20 @@ The <dfn attribute for="XRProjectionLayer">ignoreDepthValues</dfn> attribute, if
...
+XRTextureType {#xrtexturetype}
+-------------
+The {{XRTextureType}} enum defines what type of texture is allocated.
+
+<pre class="idl">
+enum XRTextureType {
+  "texture",
+  "texture-array"
+};
+</pre>
```
toji, line 250: I have a very slight preference for re-using what WebGL already defined for us, but I can see that I'm in the minority reading through #39 and this _does_ make the validation simpler, so 🤷‍♀.

webxrlayers-1.bs
```diff
@@ -479,13 +492,13 @@ interface XRWebGLLayerFactory {
 
   readonly attribute double nativeProjectionScaleFactor;
 
-  Promise&lt;XRProjectionLayer&gt; requestProjectionLayer(GLenum textureTarget, 
+  Promise&lt;XRProjectionLayer&gt; requestProjectionLayer(XRTextureType textureTarget,
```
toji, line 495: These arguments should maybe be renamed to `textureType` now? "Target" is a very WebGL-centrix term that I only used originally to help better suggest which of the wide array of GL enums should be used here.

cabanier: I will make that change

Issue #38, opened 2020-03-31
Can the XR___LayerFactory names be more general purpose?


toji:
I've suggested this before, but felt at the time it was more important to get some of the basic usage patterns nailed down first. Now that we've established the broad strokes of the API and I've got a concrete use case outside of layers to point at as an example I'd like to revisit it.

In the Lighting Estimation API we need to deliver a cube map in an API-appropriate texture format. In order to make the WebXR/WebGL/(Eventual WebGPU) interaction easier the [current explainer IDL](https://github.com/immersive-web/lighting-estimation/blob/master/lighting-estimation-explainer.md#appendix-a-proposed-partial-idl) proposes using the `XRWebGLLayerFactory` as the method of surfacing that texture, since it's already well positioned to serve up WebGL-specific resources for the session. 

The only real issue with this route is that it makes the name `XRWebGLLayerFactory` a misnomer, since it would provide more than just layers. [@kearwood and I would both prefer](https://github.com/immersive-web/lighting-estimation/pull/26#issuecomment-604656806) to see the name changes to something more generic as a result.

I'm really flexible on what that name is, as long as it suggests that it handles more WebGL resources than just layers. A few suggestions in no particular order of preference.

- `XRWebGLInterface` (may be confused with the IDL "interface" concept)
- `XRWebGLBinding`
- `XRWebGLGraphicsBinding`
- `XRWebGLSessionBinding`
- `XRWebGLResourceProvider`

cabanier:
We could drop the "layer" part and rename it to `XRWebGLFactory`.
Otherwise, I have a very slight preference to `XRWebGLInterface` and `XRWebGLBinding `

RafaelCintron:
Reading through the layers proposal, I am having a hard time understanding why we need the concept of "factories" or "bindings" in the first place.  

Instead of:

````JS
const gl = canvas.getContext('webgl', { xrCompatible: true });
const glLayerFactory = new XRWebGLLayerFactory(xrSession, gl);
const layer = await glLayerFactory.requestProjectionLayer(gl.TEXTURE_2D, { alpha: false });
````

why can't we have?

````JS
const gl = canvas.getContext('webgl', { xrCompatible: true });
const layer = await xrSession.requestWebGLProjectionLayer(gl, gl.TEXTURE_2D, { alpha: false });
````

The second version is less code and still feature detectable by web developers. 

[...]

PR #53, opened 2020-04-07
rename XRWebGLLayerFactory to XRWebGLBinding

cabanier:


cabanier:
This will resolve #38

webxrlayers-1.bs
```diff
@@ -147,8 +147,8 @@ Application flow {#applicationflow}
 
 If an author wants to use layers, they have to go through these steps:
  1. request support through {{XRPermissionDescriptor/requiredFeatures}} or {{XRPermissionDescriptor/optionalFeatures}} in {{XRSystem/requestSession()}}
- 1. create a XRWebGLLayerFactory or XRMediaLayerFactory
- 1. create layers with these factories
+ 1. create a {{XRWebGLBinding}} or {{XRMediaLayerFactory}}
```
toji, line 150: Should we also rename `XRMediaLayerFactory` -> `XRMediaBinding`? It's a bit different in that it's not really "binding" to anything at creation time (except maybe the DOM?) but it seems like it would be ideal to use the same terminology for the same basic concept of supplying imagery to/from the session.

cabanier: yes. I will make that change.

Issue #32, opened 2020-03-12
Should the requestxxx APIs use promises?


cabanier:
Currently the APIs to create layers all return a promise.
This means that the promise for the layers is fulfilled outside a Raf. I think this is confusing because the author won't be allowed to use the layers there.
Moreover, XRWebGLLayer can just be constructed so why can't we have immediate layer creation?

I think we should get rid of the promises and return the layer object directly.

toji:
I don't object to allowing direct return of a (nullable) object if we know with reasonable certainty that UAs won't need to do a round trip check to the runtime to verify whether or not the layer can be created as requested. I think that should be feasible, but we'll want to check.

In terms of using a layer inside a rAF call, I don't think that would work even if the object returned immediately? The layers are set via updateRenderState, which intentionally defers applying changes made in the rAF call until the next rAF. This is so that the compositor can be confident about the state being used for the frame in the case that the runtime needs to do some prep for it. In this specific case, I can imagine that it would be useful to know which layers are going to be used during the frame so that you only have to get swapchain images for the ones that are actually being drawn. If the active layers can change during the rAF callback then the UA can't do any work to prepare the layer's to be composited in advance or, conversely, has to do work for every layer, whether it's in use or not, on the off chance that it _might_ be used.

cabanier:
/agenda let's discuss if we need the promises

Manishearth:
>  I think this is confusing because the author won't be allowed to use the layers there.

You need to `updateRenderState()` first so it's better if these are exposed outside of the raf.

cabanier:
During today's meeting, nobody objected to removing the promise and returning the object.
I will the needed changes to the spec.

PR #54, opened 2020-04-07
removed promises from XRWebGLLayerFactory and XRMediaLayerFactory

cabanier:


cabanier:
This will resolve #32

webxrlayers-1.bs
```diff
@@ -492,18 +492,18 @@ interface XRWebGLLayerFactory {
...
-                                       XRLayerInit init);
-  Promise&lt;XRCylinderLayer&gt; requestCylinderLayer(XRTextureType textureType,
-                                               XRLayerInit init);
-  Promise&lt;XREquirectLayer&gt; requestEquirectLayer(XRTextureType textureType,
-                                               XRLayerInit init);
-  Promise&lt;XRCubeLayer&gt; requestCubeLayer(XRLayerInit init);
-
-  XRWebGLSubImage? getSubImage(XRLayer layer, XRFrame frame);
-  XRWebGLSubImage? getViewSubImage(XRLayer layer, XRView view);
+  XRProjectionLayer requestProjectionLayer(XRTextureType textureType,
```
toji, line 495: Ooh! One late comment! We've been using the term "request" exclusively with methods that return Promises till now. I wonder if this change means these should be called "create____Layer" instead?

cabanier:
> LGTM to remove the Promises. I question whether or not throwing here is the right pattern, as it seems like it could catch some developers by surprise, but I'm really not sure what the common pattern is across other web APIs for this kind of behavior.

Yes. I was wondering about that as well. One nice thing about exceptions is that we can give more information but we could also log the reason in the console.

> I'm gonna reach out to some web ergonomics folks on the Google side to get some more information, but I'm fine with this landing in the meantime because the delta between throwing an exception and returning null is smaller (from an algorithm standpoint) than returning a promise vs not.

thanks!

webxrlayers-1.bs
cabanier: I like "create" better than "request". 
I will make that change.

PR #63, opened 2020-04-09
cleaned up getViewSubImage and getSubImage text + fixed some bikeshed warnings

cabanier:


webxrlayers-1.bs
```diff
@@ -28,10 +32,13 @@ spec: WebXR Device API - Level 1; urlPrefix: https://www.w3.org/TR/webxr/#
     type: dfn; text: native WebGL framebuffer resolution
     type: dfn; text: immersive session
     type: dfn; text: xr compatible
-    type: dfn; text: "xr device"
     type: dfn; text: ended
     type: dfn; text: context
     type: dfn; text: session
+    type: dfn; text: XRFrame/active
```
Manishearth, line 38: all of these should just be link-defaults

cabanier: OK. I will address that when I fix all warnings

Issue #24, opened 2020-02-27
new interface for `updateRenderState'


cabanier:
We want to pass in an array of layers into XRRenderStateInit.
What would be the best way to express this in IDL?

Should we extend the existing dictionary or should we create a new one?
```
dictionary XRRenderStateInit {
  double depthNear;
  double depthFar;
  double inlineVerticalFieldOfView;
  XRWebGLLayer? baseLayer;
  FrozenArray<XRWebGLLayer or XRLayer> layers;
}; 
```

@Manishearth what are your thoughts?

Manishearth:
So WebIDL _does_ have `partial dictionary`, but it seems like it's generally considered a big mistake.

We have two options. We define `dictionary XRRenderStateInitLayers: XRRenderState` and in the layers spec "patch" `updateRenderState()` to accept this argument (this isn't exactly kosher iirc). Alternatively, we define a blank `XRLayer` interface in the core spec, `partial interface` it in the layers spec if necessary, and define a semantics-free `layers` array in the core spec.

cabanier:
> So WebIDL _does_ have `partial dictionary`, but it seems like it's generally considered a big mistake.

A quick search in the Chrome codebase shows

> We have two options. We define `dictionary XRRenderStateInitLayers: XRRenderState` and in the layers spec "patch" `updateRenderState()` to accept this argument (this isn't exactly kosher iirc). Alternatively, we define a blank `XRLayer` interface in the core spec, `partial interface` it in the layers spec if necessary, and define a semantics-free `layers` array in the core spec.
> 
> I'd prefer the latter.

cabanier:
> So WebIDL _does_ have `partial dictionary`, but it seems like it's generally considered a big mistake.

I find no current `partial dictionary` web interfaces in the chromium code base.
I agree that we shouldn't be the first one to add it.

> We have two options. We define `dictionary XRRenderStateInitLayers: XRRenderState` and in the layers spec "patch" `updateRenderState()` to accept this argument (this isn't exactly kosher iirc). Alternatively, we define a blank `XRLayer` interface in the core spec, `partial interface` it in the layers spec if necessary, and define a semantics-free `layers` array in the core spec.
> 
> I'd prefer the latter.

Yes and this change is very unlikely to break sites.
Can we still make normative changes to the WebXR spec?

avaerkazmer:
There are rendering schemes that use a different depth discipline, such as Unity. Depth near and far doesn’t fully specify such cases.

[...]

PR #64, opened 2020-04-09
Setting the layers array causes baseLayer to report null

asajeffrey:
Calling `updateRenderState()` with `layers` causes `baseLayer` to be `null`.


asajeffrey:
See discussion at #24

cabanier:
That looks good to me

cabanier:
I'd like @toji's input before merging this.

PR #67, opened 2020-04-10
Fix for determining texture layout

cabanier:


asajeffrey:
Thanks! That makes more sense to me!

PR #70, opened 2020-04-10
initial changes to updateRenderState

cabanier:
This is a first pass at changing existing WebXR behavior for layers.
I'm unsure if this is the correct way to so feedback is appreciated :-)

I'm not planning on submitting this in the current state.

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -23,6 +23,9 @@ spec: webxr-1;
     type: dfn; text: xr device; for: /
     type: dfn; text: active; for: XRFrame
     type: dfn; text: eye; for: XRView
+    type: dfn; text: initialize the render state
+    type: dfn; text: pending render state
+    type: dfn; text: apply the pending render state
 </pre>
 
 <pre class="anchors">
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -797,6 +800,69 @@ interface XRMediaBinding {
...
+
+The <dfn dict-member for="XRRenderStateInit">layers</dfn> is an array containing
+instances of {{XRLayer}} or {{XRWebGLLayer}} to hand to the [=XR Compositor=].
+
+The <dfn attribute for="XRRenderState">layers</dfn> attribute returns an array containing
+the instances of {{XRLayer}} or {{XRWebGLLayer}} that are displayed by the [=XR Compositor=].
+
+<div class="algorithm" data-algorithm="initialize-renderstate-for-layers">
+When an {{XRRenderState}} object |state| is created for an {{XRSession}} |session| with {{layers}} enabled, the user agent MUST <dfn dfn-for="layers">initialize the render state</dfn> by running the following steps:
+  1. Initialize |state| by running the steps to [=initialize the render state=].
```
toji, line 839: Is this self referential to the dfn in the previous line? I think the structure here is good, I just want to make sure the linking is working as intended.

webxrlayers-1.bs
```diff
@@ -797,6 +800,69 @@ interface XRMediaBinding {
...
+  FrozenArray&lt;XRGenericLayer&gt;? layers;
+};
+
+[SecureContext, Exposed=Window] interface XRRenderState {
+  readonly attribute double depthNear;
+  readonly attribute double depthFar;
+  readonly attribute double? inlineVerticalFieldOfView;
+  readonly attribute XRWebGLLayer? baseLayer;
+  readonly attribute FrozenArray&lt;XRGenericLayer&gt;? layers;
+};
```
toji, line 828: This can be a partial interface, and I think we should define it as such to highlight what this module is actually adding.

webxrlayers-1.bs
```diff
@@ -797,6 +800,69 @@ interface XRMediaBinding {
...
+<pre class="idl">
+typedef (XRWebGLLayer or XRLayer) XRGenericLayer;
+
+dictionary XRRenderStateInit {
+  double depthNear;
+  double depthFar;
+  double inlineVerticalFieldOfView;
+  XRWebGLLayer? baseLayer;
+  FrozenArray&lt;XRGenericLayer&gt;? layers;
+};
```
toji, line 820: Yeah... this feels awkward but all things considered it's probably the best way to define this dictionary, given that we can't do partials. In any case, I think it's a good place to start as it clearly shows our intent, and then we can haggle it out with the IDL experts later.

webxrlayers-1.bs
```diff
@@ -797,6 +800,69 @@ interface XRMediaBinding {
...
+</div>
+
+updateRenderState changes {#updaterenderstatechanges}
+-------------------------
+<div class="algorithm" data-algorithm="updateRenderState-for-layers">
+When the <dfn method for="XRSession" dfn-for="layers">updateRenderState(|newState|)</dfn> method is called on an {{XRSession}} |session|
+with {{layers}} enabled, the user agent MUST run the following steps:
+
+  1. Run {{XRSession/updateRenderState}} with |newState|.
+  1. If |newState|'s {{XRRenderStateInit/baseLayer}} is set, set |session|'s [=pending render state=] to <code>null</code>, throw an {{InvalidStateError}} and abort these steps.
```
toji, line 851: Ah, I didn't remember where we had landed on this. Just to double check: The intent is that if we enabled "layers" in the optional/requiredFeatures then `baseLayer` is disabled entirely, right? I think that's fine, but we'll need to ensure pages have a way to detect that layers did actually get enabled in the case that they passed it as an optional feature, so that the developers know which render state to set.

webxrlayers-1.bs
cabanier: No, it refers to the definition in the WebXR spec. 
The one in the previous line has a `dfn-for="layers"` so it's different.

webxrlayers-1.bs
cabanier: Will do.

webxrlayers-1.bs
cabanier: > Ah, I didn't remember where we had landed on this. 

Yes, I think we discussed it but it's not clear from the readme what we had decided.

> Just to double check: The intent is that if we enabled "layers" in the optional/requiredFeatures then `baseLayer` is disabled entirely, right? 

Correct

> I think that's fine, but we'll need to ensure pages have a way to detect that layers did actually get enabled in the case that they passed it as an optional feature, so that the developers know which render state to set.

The way I tried to structure the spec is that you will get an empty layers array in the XRRenderState (as opposed to it being absent).

webxrlayers-1.bs
cabanier: It creates a link to the steps in WebXR.

Manishearth:
(apologies for not having time to respond to your comments in the issue earlier, I'll try to help with the design here asap!)

webxrlayers-1.bs
toji: > The way I tried to structure the spec is that you will get an empty layers array in the XRRenderState (as opposed to it being absent).

Excellent, thanks!

webxrlayers-1.bs
```diff
@@ -797,6 +800,69 @@ interface XRMediaBinding {
...
+
+<pre class="idl">
+typedef (XRWebGLLayer or XRLayer) XRGenericLayer;
+
+dictionary XRRenderStateInit {
+  double depthNear;
+  double depthFar;
+  double inlineVerticalFieldOfView;
+  XRWebGLLayer? baseLayer;
+  FrozenArray&lt;XRGenericLayer&gt;? layers;
```
Manishearth, line 819: this should be `sequence`, as an input type

webxrlayers-1.bs
```diff
@@ -797,6 +800,69 @@ interface XRMediaBinding {
...
+
+XRRenderState changes {#xrrenderstatechanges}
+---------------------
+This module extends the {{XRRenderStateInit}} and {{XRRenderState}} interfaces with a new optional array {{XRRenderStateInit/layers}} containing
+instances of {{XRLayer}} or {{XRWebGLLayer}}.
+
+<pre class="idl">
+typedef (XRWebGLLayer or XRLayer) XRGenericLayer;
+
+dictionary XRRenderStateInit {
```
Manishearth, line 814: Eventually these changes should be upstream, webidl is not a form of such patching

I've written this in more detail in my comment https://github.com/immersive-web/layers/issues/24#issuecomment-612966939

webxrlayers-1.bs
```diff
@@ -797,6 +800,69 @@ interface XRMediaBinding {
...
+  1. For each |layer| in |newState|'s {{XRRenderStateInit/layers}}:
+    1. If |layer|'s [=XRLayer/session=] is different from |session|, throw an {{InvalidStateError}} and abort these steps.
+  1. If |newState|'s {{XRRenderStateInit/layers}} array is set, set |session|'s [=pending render state=]'s {{XRRenderState/layers}} to |newState|'s {{XRRenderStateInit/layers}}.
+
+</div>
+
+<!--
+<div class="algorithm" data-algorithm="apply-pending-render-state-for-layers">
+
+When the {{XRSession}} MUST <dfn>apply the pending render state for an {{XRSession}} |session| with {{layers}} enabled</dfn>, the user agent MUST run the following steps:
```
Manishearth, line 861: we also need to null out the base layer, or set the base layer to the first layer

(also, are there layer types that should not be the first layer? must the first layer only be XRWebGLLayer or XRProjectionLayer?)

cabanier: > we also need to null out the base layer, or set the base layer to the first layer

I don't believe that is needed because we set the pending render state to null if the base layer is set.

> (also, are there layer types that should not be the first layer? must the first layer only be XRWebGLLayer or XRProjectionLayer?)

No, there are no restrictions on what layer is first.

PR #80, opened 2020-04-16
Remove transform

cabanier:
removed RigidTransform + added referenceSpace to init dictionary


cabanier:
I will hold off from submitting this for now.

PR #84, opened 2020-04-16
Added XRSpace validation

cabanier:


webxrlayers-1.bs
```diff
@@ -261,6 +261,14 @@ of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.
 Each {{XRLayer}} has an associated <dfn for="XRLayer">session</dfn>, which is the
 {{XRSession}} it was created with.
 
+<div class="algorithm" data-algorithm="setting the space on a layer">
```
Manishearth, line 264: Should the space field just be on `XRLayer` and consolidate? we can use `partial interface` here.

cabanier: Equirect and projection layers unfortunately don't have a `space` attribute.

Manishearth: Ah, I missed that. For some reason I thought projection layers aren't `XRLayer`s, but it's that they don't use <code>XLayer<i>Init</i></code>. 

We could _potentially_ have an `XRPlaceableLayer` interface that sits in between, but I'm not convinced that's super useful.

webxrlayers-1.bs
```diff
@@ -298,6 +306,7 @@ interface XRQuadLayer : XRLayer {
 The <dfn attribute for="XRQuadLayer">transform</dfn> attributes sets and returns the offset and orientation relative to the
 <dfn attribute for="XRQuadLayer">space</dfn> attribute. The {{XRQuadLayer/transform}} and {{XRQuadLayer/space}} attributes
 establish the spatial relationship of the layer within the user’s physical environment.
+When setting the {{XRQuadLayer/space}}, first run the steps for [=setting the space on a layer=].
```
Manishearth, line 309: _Technically_ all of these attributes should be reflecting internal slots that we define, but that's nbd (spec review might catch this eventually). I might make that change later.

webxrlayers-1.bs
cabanier: Can we define it in IDL so it's just a helper class (ie you can't construct it)

webxrlayers-1.bs
cabanier: Is that done in the WebXR spec? If so, I can follow that pattern

webxrlayers-1.bs
Manishearth: You can't construct IDL interfaces by default (this is true of XRLayer too)

Manishearth: I'm mostly proposing this because I think it will simplify the spec text, but I'm not sure (and i'm not terribly attached to it).

cabanier: o yes, but there's still `instanceof`

Manishearth: There's NoInterfaceObject, but we shouldn't do that. `instanceof` handles inheritance just fine.

The reason behind proposing this was not to make it easier to use from JS (it changes nothing), just to make it easier to consolidate spec text.

PR #86, opened 2020-04-17
Fix incorrect usage of WebGL2RenderingContext

cabanier:


cabanier:
> Shouldn't WebGL2RenderingContext still be specified for texture-arrays?

you're right. This should be XRWebGLRenderingContext.
I will fix this in my repo first.

PR #88, opened 2020-04-17
Refactor XRLayers

cabanier:


webxrlayers-1.bs
```diff
@@ -221,15 +224,15 @@ enum XRLayerLayout {
...
-Every layer except XRWebGLLayer is derived from XRLayer.
-XRLayer defines a set of common attributes and behaviors acress all layer types.
+XRGenericLayer {#xrgenericbaselayertype}
+--------------
+Every layer except XRWebGLLayer is derived from XRGenericLayer.
+XRGenericLayer defines a set of common attributes and behaviors acress all layer types.
 
 <pre class="idl">
-interface XRLayer {
+interface XRGenericLayer {
```
Manishearth, line 235: I don't like this name, but that's a very minor concern.

webxrlayers-1.bs
```diff
@@ -954,9 +940,10 @@ When the <dfn method for="XRSession" dfn-for="layers">updateRenderState(|newStat
 with "[=feature descriptor/layers=]" enabled, the user agent MUST run the following steps:
```
Manishearth, line 940: This needs to be updated to patch the `update the pending layers state` algorithm.

Something like, "When the `update the pending layers state` algorithm needs to be called, instead call the `update the pending layers state with layers support` algorithm, defined below.

The algorithm should do a feature check, and also do all the other stuff below

cabanier: > This needs to be updated to patch the `update the pending layers state` algorithm.

That is what this code is doing.

> Something like, "When the `update the pending layers state` algorithm needs to be called, instead call the `update the pending layers state with layers support` algorithm, defined below.

Are you saying that it needs to be more explicit that it's replacing the other algorithm?

> The algorithm should do a feature check, and also do all the other stuff below

Isn't the feature check covered by "an XRSession session with "layers" enabled"?
What do you mean by "other stuff below"? All that needs to happen is to have an empty array.

Manishearth: Right, the reason we introduced the hook is so that we don't have to replace the entire updateRenderState algorithm, just this part.

It's fine to also make the patch dynamic and dependent on the feature, but it's better if that's in the spec text IMO so that we can clearly define how the behavior is with and without the feature.

cabanier: OK. I just updated the patch and I think it's better. Please take a look and thanks for all your help!

webxrlayers-1.bs
```diff
@@ -912,63 +915,50 @@ WebXR Device API Integration {#webxrintegration}
...
 
-  1. Run {{XRSession/updateRenderState}} with |newState|.
-  1. If |newState|'s [=XRRenderStateInit/baseLayer=] is set, set |session|'s [=pending render state=] to <code>null</code>, throw an {{InvalidStateError}} and abort these steps.
+<div class="algorithm" data-algorithm="update-layers-state-for-layers">
+
+This module replaces the steps given by "[=update the pending layers state=]". Instead when the user agent will
+<dfn dfn-for="layers">update the pending layers state</dfn> with {{XRSession}} |session| and {{XRRenderStateInit}} |newState|, it must run the following steps:
+  1. If the |session| was not created with "[=feature descriptor/layers=]" enabled:
+    1. If {{XRRenderStateInit/layers}} is not <code>null</code>, throw an {{NotSupportedError}} and abort these steps.
+    1. Abort these steps.
```
Manishearth, line 952: you're aborting twice here, pick one option :smile:

webxrlayers-1.bs
```diff
@@ -912,63 +915,50 @@ WebXR Device API Integration {#webxrintegration}
...
+  1. If |newState|'s {{XRRenderStateInit/baseLayer}} is set, throw a {{InvalidStateError}} and abort these steps.
```
Manishearth, line 953: These steps are now called unconditionally, so this should check for both layers and baselayer being set

webxrlayers-1.bs
```diff
@@ -912,63 +915,50 @@ WebXR Device API Integration {#webxrintegration}
...
+  1. If |newState|'s {{XRRenderStateInit/layers}} contains duplicate instances, throw an {{InvalidStateError}} and abort these steps.
   1. For each |layer| in |newState|'s {{XRRenderStateInit/layers}}:
-    1. If |layer|'s [=XRLayer/session=] is different from |session|, throw an {{InvalidStateError}} and abort these steps.
+    1. If |layer| is a {{XRGenericLayer}} and |layer|'s [=XRGenericLayer/session=] is different from |session|, throw an {{InvalidStateError}} and abort these steps.
+    1. If |layer| is a {{XRWebGLLayer}} and |layer|'s [=XRWebGLLayer/session=] is different from |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If |newState|'s {{XRRenderStateInit/layers}} array is set, set |session|'s [=pending render state=]'s {{XRRenderState/layers}} to |newState|'s {{XRRenderStateInit/layers}}.
```
Manishearth, line 958: This should also set `pending/baseLayer` to null, and we should have another line that nulls out `pending/layers` when `baseLayer` is set.

webxrlayers-1.bs
cabanier: Are you saying `layers` shouldn't be null?
Like: If baseLayers is set or layers is null, throw InvalidStateError

I think that'd be wrong.

webxrlayers-1.bs
cabanier: done!

webxrlayers-1.bs
cabanier: Are you saying layers shouldn't be null?
Like: If baseLayers is set or layers is null, throw InvalidStateError

I think that would be wrong.

webxrlayers-1.bs
cabanier: Are you saying layers shouldn't be null?
Like: If baseLayers is set or layers is null, throw InvalidStateError

I think that would be wrong.

webxrlayers-1.bs
cabanier: Why is that needed?

cabanier:
@Manishearth I moved the check to earlier. Otherwise the `pending render state` was updated.
I also updated the layers spec to deal with this.

webxrlayers-1.bs
Manishearth: Otherwise you can do `updateRenderState(baseLayer: foo)` followed by `updateRenderState(layers: bar)`. We either need to silently null out `baseLayer` when `layers` is set and vice versa, or we need to have code that ensure that the active render state also does not have baseLayer or layers set when we are trying to update the other without explicitly nulling out the first.

cabanier: ah. But you can't do that anyway :-)
If the layers feature isn't set, you will error out if you pass a layers array
Likewise, if it is set, you will get an error if you pass a baseLayer.

Manishearth: oh! I didn't realize that, that's clever

I don't see where in the algorithm we error if baseLayer is used with the feature set, though?

And this doesn't handle cases where the feature set is appended midway

cabanier: > I don't see where in the algorithm we error if baseLayer is used with the feature set, though?

See line 954. It's hard to see on the review.

> And this doesn't handle cases where the feature set is appended midway

That can't happen. The feature is requested when the session is created.

Manishearth: > That can't happen. The feature is requested when the session is created.

No, with the permissions API integration (I don't think this is shipped anywhere yet, but it's there), you can request features at arbitrary points

https://immersive-web.github.io/webxr/#ref-for-xr-device-list-of-enabled-features%E2%91%A0

You could declare the layers feature as one which _has_ to be requested at session creation. @toji ?

webxrlayers-1.bs
Manishearth: I don't think I understand you, what I'm saying is that the new "update the pending layers state" algorithm is unconditionally called, even when the feature is not set (this is good).

This should only error in cases when layers are being used -- either by checking the feature as discussed below, or when pending/active's layers is also set.

We should just give this algorithm two branches for when layers is enabled vs not enabled.

cabanier: That is what it is doing. Line 951 is true if layers isn't requested.

webxrlayers-1.bs
Manishearth: Actually, I was wrong here, aborting twice is okay

webxrlayers-1.bs
Manishearth: Oh! I misread that, never mind!

webxrlayers-1.bs
cabanier: No big deal 👍

webxrlayers-1.bs
cabanier: > You could declare the layers feature as one which _has_ to be requested at session creation. @toji ?

I think it doesn't make sense to add this at runtime because an author can use the classic welgllayer as the sole entry in the array. Granting this permission will allso put the renderState in a bad state.

Manishearth: > Granting this permission will allso put the renderState in a bad state.

Right, that's why I was initially suggesting the additional check.

We should specify somewhere that the "layers" feature cannot be requested after session creation.

webxrlayers-1.bs
toji: One other option, I guess, is that since the amount of shared values here is limited and there's certain attributes that we're sharing across all but one layer type already, maybe we just skip having this intermediate layer type all together and have each layer copy these into their own interface.

It's not the most elegant thing ever, but there's not a whole lot that this structure concretely buys us anyway besides preventing you from passing an `XRWebGLLayer` into `get[View]SubImage()`. But even that will need to make sure that the layer was not created by a different `XR____Binding` object in it's algorithm anyway (because we don't want WebGL bindings to return images for WebGPU-created layers.) So maybe i wouldn't be a terrible things to drop this type from the hierarchy.

webxrlayers-1.bs
toji: Hrm. I forgot about the permissions API's mid-session feature request capability.

I do think there are going to be features that don't make sense to request once the session has begun, largely if the underlying runtime requires that the feature be enabled at startup. In this particular I don't think that restriction applies, but I also can't think of any reason why the page would want to switch layering style mid-session, so I woud have no problem with putting that restriction in place. I don't know what the best way to flag that would be, though.

webxrlayers-1.bs
cabanier: > If we end up keeping `XRGenericLayer` it's fine, it just sounds slightly awkward to me.

I like the name `XRCompositionLayer`. I will update the PR with that name.

cabanier: > One other option, I guess, is that since the amount of shared values here is limited and there's certain attributes that we're sharing across all but one layer type already, maybe we just skip having this intermediate layer type all together and have each layer copy these into their own interface.

There's still quite a bit of common interface between the composition layers so I think we should keep a common base class.

webxrlayers-1.bs
cabanier: I will file an issue on this so we're not gated on solving it in the PR.

PR #91, opened 2020-04-21
Added cubemap-texture type + switched to DOMPointReadOnly

cabanier:
This PR introduces `cubemap-texture`.
More work is needed to clarify how cubemaps work.

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -592,6 +592,7 @@ MUST be multiplied by to yield the |session|'s [=native WebGL framebuffer resolu
 <div class="algorithm" data-algorithm="determine the layout attribute">
 
 To <dfn>determine the layout attribute</dfn> using a {{XRTextureType}} |textureType|, a {{XRWebGLRenderingContext}} |context| and a {{XRLayerInit}} |init|, the user agent MUST run the following steps:
+  1. If |textureType| is {{"cubemap-texture"}}, throw {{InvalidStateError}} and abort these steps.
   1. If |context| is not a {{WebGL2RenderingContext}} and |textureType| is {{"texture-array"}}, throw {{InvalidStateError}} and abort these steps.
   1. let |layout| be |init|'s {{XRLayerInit/layout}}.
   1. If |layout| is {{XRLayerLayout/"mono"}}, return |layout| and abort these steps.
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -447,14 +447,12 @@ The {{XRTextureType}} enum defines what type of texture is allocated.
 <pre class="idl">
 enum XRTextureType {
   "texture",
-  "texture-array",
-  "cubemap-texture"
+  "texture-array"
 };
 </pre>
 
 - A texture type of <dfn enum-value for="XRTextureType">texture</dfn> indicates that the textures of {{XRWebGLSubImage}} must be of type {{TEXTURE_2D}}
 - A texture type of <dfn enum-value for="XRTextureType">texture-array</dfn> indicates that the textures of {{XRWebGLSubImage}} must be of type {{TEXTURE_2D_ARRAY}}
-- A texture type of <dfn enum-value for="XRTextureType">cubemap-texture/dfn> indicates that the textures of {{XRWebGLSubImage}} must be of type {{TEXTURE_CUBE_MAP}}
 
 GPU layer and view creation {#gpulayer}
 ===========================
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -446,13 +447,14 @@ The {{XRTextureType}} enum defines what type of texture is allocated.
 <pre class="idl">
 enum XRTextureType {
   "texture",
-  "texture-array"
+  "texture-array",
+  "cubemap-texture"
```
Artyom17, line 451: I am not sure we need publicly expose "cubemap-texture". The cube layer may use only that, but we do not specify it the creation parameters.

cabanier: You're right. I mistakenly thought it was an attribute. 
I will fix.

webxrlayers-1.bs
```diff
@@ -797,7 +807,7 @@ When this method is invoked, the user agent MUST run the following steps:
   1. Initialize |layer|'s [=XRCompositionLayer/session=] to |session|.
   1. Let |layer|'s {{XRCubeLayer/space}} be the {{session}}'s [=viewer reference space=].
   1. Let |layer|'s {{XRCubeLayer/orientation}} be a new {{DOMPoint}}.
-  1. let |textureType| be {{XRTextureType/"texture-array"}}.
+  1. let |textureType| be {{XRTextureType/"cubemap-texture"}}.
```
Artyom17, line 811: We do not have textureType exposed to user, right?

toji: No, I just double checked that myself. We shouldn't have this as a publically accessible value, given that it's never actually going to be exposed by the API and it's use is implied by the layer type. The DOMPointReadOnly change is good, though.

webxrlayers-1.bs
```diff
@@ -796,18 +797,27 @@ When this method is invoked, the user agent MUST run the following steps:
...
   1. Initialize |layer|'s [=XRCompositionLayer/session=] to |session|.
   1. Let |layer|'s {{XRCubeLayer/space}} be the {{session}}'s [=viewer reference space=].
-  1. Let |layer|'s {{XRCubeLayer/orientation}} be a new {{DOMPoint}}.
-  1. let |textureType| be {{XRTextureType/"texture-array"}}.
+  1. Let |layer|'s {{XRCubeLayer/orientation}} be a new {{DOMPointReadOnly}}.
   1. let |layout| be |init|'s {{XRLayerInit/layout}}.
   1. If |layout| is {{XRLayerLayout/"stereo-left-right"}} or {{XRLayerLayout/"stereo-top-bottom"}}, throw {{InvalidStateError}} and abort these steps.
-  1. let |layer|'s [=colorTextures=] be the result of [=allocate color textures|allocating color textures=] with |layout|, |textureType|, |context| and |init|.
-  1. let |layer|'s [=depthStencilTextures=] be the result of [=allocate depth textures|allocating depth textures=] with |layout|, |textureType|, |context| and |init|.
+  1. Initialize |array| as follows:
```
Manishearth, line 803:
```suggestion
  1. Initialize |array| as follows, based on the value of |layout|:
```

webxrlayers-1.bs
```diff
@@ -611,17 +611,18 @@ To <dfn>allocate color textures</dfn> using a {{XRLayerInit/layout}} |layout|, a
   1. If |layout| is {{XRLayerLayout/mono}}:
         <dl class="switch">
         <dt> If |textureType| is {{"texture-array"}}:
-        <dd> Initialize |array| with 1 new instance of {{WebGLTexture}} created as a {{TEXTURE_2D_ARRAY}} texture with 1 internal texture using |context| and |init|'s {{XRLayerInit/alpha}}, {{XRLayerInit/pixelWidth}} and {{XRLayerInit/pixelHeight}} values.
+          <dd> Initialize |array| with 1 new instance of {{WebGLTexture}} created as a {{TEXTURE_2D_ARRAY}} texture with 1 internal texture using |context| and |init|'s {{XRLayerInit/alpha}}, {{XRLayerInit/pixelWidth}} and {{XRLayerInit/pixelHeight}} values.
         <dt> Otherwise
-        <dd> Initialize |array| with 1 new instance of {{WebGLTexture}} created as a {{TEXTURE_2D}} texture with |context| and |init|'s {{XRLayerInit/alpha}}, {{XRLayerInit/pixelWidth}} and {{XRLayerInit/pixelHeight}} values.
+          <dd> Initialize |array| with 1 new instance of {{WebGLTexture}} created as a {{TEXTURE_2D}} texture with |context| and |init|'s {{XRLayerInit/alpha}}, {{XRLayerInit/pixelWidth}} and {{XRLayerInit/pixelHeight}} values.
+        <dl>
```
Manishearth, line 617: is there a reason why the cube map case isn't handled differently here? just curious

cabanier: It turned out pretty messy.
Also, cubemap will require different calculations for width and height which will make it more verbose.

PR #100, opened 2020-04-23
Added language for immutable texture

### LINKED ISSUE #96: Clarify that textures from layers are [most likely] immutable
Since textures from the layers are pre-allocated they most likely are "immutable" (see https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf#nameddest=section-3.8.4). There are several consequences to that which affect how dev deal with such textures:
* all texImage* functions will fail with INVALID_OPERATION; texSubImage* must be used instead.
* copyTexImage* functions won't work either; copySubTexImage must be used.

Note, it doesn't matter if those textures are used in WebGL 1 or 2; it is a matter how those textures were allocated (via glTexStorage*, which is supported in GL ES 3.x / OpenGL 4.2, or via ARB_texture_storage extension for OpenGL ES 1.0, OpenGL ES 2.0, OpenGL 1.2 (https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_storage.txt)).

Probably, we would need to add "non-normative" section about it somewhere.

cabanier:
Fixes #96

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -441,8 +441,8 @@ If the layer was created without depth/stencil, this attribute returns null.
 The <dfn attribute for="XRWebGLSubImage">imageIndex</dfn> attribute returns the offset into the texture array. Valid only for layers
 that were requested with {{texture-array}}.
 
-{{XRWebGLSubImage/colorTexture|ColorTexture}} and {{XRWebGLSubImage/depthStencilTexture}} are only valid during the
-{{XRSession/requestAnimationFrame()}} callback and becomes invalid as soon as the callback returns.
+The {{XRWebGLSubImage/colorTexture}} and {{XRWebGLSubImage/depthStencilTexture}} objects MUST only be used during the
+{{XRSession/requestAnimationFrame()}} callback of the current session and MUST be invalid as soon as the callback returns.
 
 In addition, they MUST behave as though they were allocated with [=texStorage2D=] or [=texStorage3D=], as appropriate,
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -439,8 +441,19 @@ If the layer was created without depth/stencil, this attribute returns null.
 The <dfn attribute for="XRWebGLSubImage">imageIndex</dfn> attribute returns the offset into the texture array. Valid only for layers
 that were requested with {{texture-array}}.
 
-This {{XRWebGLSubImage/colorTexture}} and {{XRWebGLSubImage/depthStencilTexture}} are only valid during
-{{XRSession/requestAnimationFrame()}} callback and becomes invalid as soon as the callback returns.
+The {{XRWebGLSubImage/colorTexture}} and {{XRWebGLSubImage/depthStencilTexture}} objects MUST only be used during the
+{{XRSession/requestAnimationFrame()}} callback of the current session and MUST be invalid as soon as the callback returns.
```
toji, line 445: Not an issue introduced by this PR, but on reading this I realized this should actually say "MUST only be used during an [=XR Animation frame=] of the current session and MUST be made invalid once the [=XR Animation frame=] completes."

The reason for this is that an animation frame may be comprised of multiple individual rAF callbacks, and we actually don't want to invalidate the texture handle after every once since the texture will only be sent to the compositor once all of them complete.

cabanier: I replaced the text with your suggestion.

Issue #90, opened 2020-04-20
Clarify that `layers` feature can't be enabled mid-session


cabanier:

cabanier:
@Manishearth you stated that the `layers` could be enabled mid-session and provided [this link](https://immersive-web.github.io/webxr/#request-the-xr-permission).
That algorithm is only used by `requestSession`. Is there another place where it is used?

Manishearth:
It's used directly by the permissions API: https://wicg.github.io/permissions-request/#dom-permissions-request

cabanier:
That one is not implemented by any browser afaik

Manishearth:
For XR? Not yet, but folks were planning to I think. Either way we should have this language in before people implement it :)

PR #102, opened 2020-04-23
Added language to disallow enabling layers during a session

cabanier:
This PR tries to fix issue #90

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -18,6 +18,8 @@ Abstract: This specification describes support for accessing the geometry of rea
 <pre class="link-defaults">
 spec:infra;
     type: dfn; text:string
+spec: permissions-request-1;
+    type: dfn; text: request(permissionDesc); for: Permissions
 spec: webxr-1;
     type: dfn; text: immersive xr device; for: XR
     type: dfn; text: xr device; for: /
@@ -38,6 +40,7 @@ spec: webxr-1;
     type: dfn; text: updateRenderState; for: XRSession
     type: dfn; text: session; for: XRWebGLLayer
     type: dfn; text: layers; for: XRRenderStateInit
+    type: dfn; text: request the xr permission
 </pre>
 
 <pre class="anchors">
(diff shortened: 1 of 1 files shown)


toji:
I'm good with this, but I'd like @mounirlamouri to take a look as our resident permissions API expert.

webxrlayers-1.bs
```diff
@@ -198,6 +201,12 @@ navigator.xr.requestSession('immersive-vr', {
 Layers are only supported for XRSessions created with XRSessionMode of {{XRSessionMode/"immersive-vr"}}
 or {{XRSessionMode/"immersive-ar"}}. {{XRSessionMode/"inline"}} sessions MUST not support layers.
 
+If a session is created without "[=feature descriptor/layers=]" support, it MUST NOT be granted the
+"[=feature descriptor/layers=]" feature using the [=request the xr permission=] algorithm.
```
Manishearth, line 205: I don't like calling out this argument in particular, can we instead just mention that the layer descriptor has a [feature requirement](https://immersive-web.github.io/webxr/#feature-requirements) that it must not be called when there is an active XRSession of the requested mode?

cabanier: > I don't like calling out this algorithm in particular, can we instead just mention that the layer descriptor has a [feature requirement](https://immersive-web.github.io/webxr/#feature-requirements) that it must not be enabled when there is an active XRSession of the requested mode, with a note saying that this cannot be enabled mid-session.

How would we specify that? The link you provided list requirement, not features that should not be enabled.

> Though with the current approach you have the ability to request this feature and let it get enabled the next session.

Yes, that's the way I wrote it since you alluded that this is possible. 
However, if you request a feature (using `request` API), will it be enabled next time you call requestSession (even if you don't enable layers)?

I think that would be really strange, and also break inline sessions since they can't have layers.

Manishearth: > How would we specify that? The link you provided list requirement, not features that should not be enabled.

sure, but the requirement could be "not in the middle of a session"

> and also break inline sessions since they can't have layers.

ah, then we can just do it for immersive sessions. it wouldn't break inline sessions though.

>  However, if you request a feature (using request API), will it be enabled next time you call requestSession (even if you don't enable layers)?

no, but it may not trigger a consent prompt this time.

the use case at hand is an application deciding later on to request layers, but that doesn't seem super useful: you should just have it in optionalFeatures initially and check later

cabanier: @Manishearth I don't quite know how to phrase this.
Can you make a suggestion?

Manishearth, line 205:
```suggestion
The [=feature descriptor/"layers"=] [=feature descriptor=] has a [=feature requirement=] that it cannot be enabled when there is an [=active immersive session=].
```

Manishearth: Done

Manishearth: Others might disagree with this approach, though :smile:, worth waiting to see what Brandon or Mounir think

cabanier: I think this is a reasonable approach. I will merge it but keep the issue open for now.

PR #110, opened 2020-04-27
Added needsRedraw attribute

### LINKED ISSUE #106: Static or rarely-updated layers: content discard notification?
This issue is related to #101 and #49, and even to old PR https://github.com/immersive-web/webxr/commit/9d3ccaf24cd202704dcd7d9c00c97b03129abef6

The situation is the following. The textures returned by the layers (via `getSubImage` or `getViewSubImage`) can be textures, directly used by the underlying VR compositor (it is so for Oculus). This way we avoid any extra copies and it is the most efficient way to render.

However, such textures might be freed / reallocated by the compositor under-the-hood. An example of such situation is when the VR device falls asleep/wakes up or when browser temporarily switches context to OS (some kind of dialog, settings, etc) and then gets back. Note, in both cases we should receive **onvisibilitychange** event. But I am not sure if there are other situations possible, which won't trigger the `onvisibilitychange` event.
The textures got re-allocated once the device wakes up or context is switches back.

It is not a problem for the layers which are updated every frame. However, it is an issue for the 'static' layers (#49) or for the layers which are not updated every frame ('semi-static'). Such layers will lose their content and they should get re-populated. 

It is not a [good] option to let UA preserve the content of such textures, because not all textures are allowed to be readout, especially if we are talking about MSAA ones.

Thus, I think, user's experience should get notification about such 'discard' event. I see two options:
* Utilize the **onvisibilitychange** and make it mandatory for the experience to handle it and to re-populate all the static and semi-static layers, or
* introduce a new event, like **ontexturesdiscard** (in session) which will target this kind of situations w/o linking it to visibility status and ask user's code to re-populate textures in this handler. This is a more universal solution, imo.

Any thoughts?

cabanier:
Closes #106

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -264,11 +264,12 @@ enable optical chromatic aberration correction for the layer. Only valid if this
 
 ISSUE: describe what chromaticAberrationCorrection does.
 
-The <dfn attribute for="XRCompositionLayer">needsRedraw</dfn> attribute signals that the {{XRCompositionLayer}} needs to be
+The <dfn attribute for="XRCompositionLayer">needsRedraw</dfn> attribute signals that the {{XRCompositionLayer}} SHOULD be
 rerendered in the next [=XR animation frame=].
 
 NOTE: the user agent might set this flag if the underlying resources were reallocated or if the [=XR Compositor=] can no
-longer reproject the layer.
+longer reproject the layer. Failing to redraw the content in the next [=XR animation frame=] might cause flickering or
+other side effects.
 
 {{destroy()}} will delete the underlying attachments. If there are no attachments, this function does nothing.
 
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -263,8 +264,24 @@ enable optical chromatic aberration correction for the layer. Only valid if this
 
 ISSUE: describe what chromaticAberrationCorrection does.
 
+The <dfn attribute for="XRCompositionLayer">needsRedraw</dfn> attribute signals that the {{XRCompositionLayer}} SHOULD be
```
toji, line 267: Really minor nit: We should only use all-caps SHOULD/MUST/MAY etc when describing UA behavior, since that's the only thing this spec can actually control. Since this describes page behavior it needs to be simply "should"

Issue #21, opened 2020-02-21
Define how input is handled


cabanier:
Only the UA really knows where layers are placed in 3D space. 
The author (or JS framework) can guess where they are but this might be brittle.

What is the best way to provide input events? Should non-projection layers become event targets?

Manishearth:
I feel like we could trigger select events with xy coordinates?

Having xy coordinates would be useful for screen-space select events as well.

cabanier:
> I feel like we could trigger select events with xy coordinates?

Would it return a x/y coordinate + layer?

Manishearth:
Oh, I was suggesting that we trigger select events _on the layer_, and we extend select events to also contain optional xy parameters when found on a layer. Or have an `XRInputSourceLayerEvent: XRInputSourceEvent` with xy coordinates. I like the former solution because we can use the same thing with screen space input.

avaer:
What about projection layers? Would that be a special UA case, or left for the page to handle?

cabanier:
> Oh, I was suggesting that we trigger select events _on the layer_, and we extend select events to also contain optional xy parameters when found on a layer. Or have an `XRInputSourceLayerEvent: XRInputSourceEvent` with xy coordinates. I like the former solution because we can use the same thing with screen space input.

I agree. I think this means that we should make layers event targets.

cabanier:
> What about projection layers? Would that be a special UA case, or left for the page to handle?

I think they could handle events like in regular WebXR,

Manishearth:
Perhaps we should open an issue on the core spec about adding xy coordinates to input source events, and perhaps extending the targetraymode to include something new for layers (we could also just use screen again).

Manishearth:
So while discussing this with @asajeffrey we realized this might be trickier than "slap an `onselect` event on `XRLayer` and call it a day".

Basically, WebXR doesn't really strongly define a pointing ray. targetRayMode is the origin of the "preferred pointing ray", but applications are free to add an offset to it (for whatever reason). So we don't have enough information to actually run a hit test, unless we provide the user a way to specify an offset for certain.

Perhaps the shape of this API needs to be similar to the hit test API, using an XRRay.

Some ideas:

 - You can register an XRRay on the input source, and it's used to trigger select events on the layer. 
 - You can create LayerHitTestSources with XRRays and input sources, and events fire on that instead.
 - The input source has an `attribute XRRay pointerRay`, initialized to `XRRay()`, that we can modify.

[...]

PR #115, opened 2020-04-30
First pass at video layers

cabanier:


cabanier:
@toji would you mind taking a quick look at this PR? It needs more work but I don't want to add too much at the same time.

Manishearth:
I haven't looked really closely at this yet (will tomorrow), but have we considered making this based on media streams? This means videos can still be piped to WebXR, but so can webrtc streams and getUserMedia.

That said, when it's a video element layer the UA can actually choose to show video controls. I'm not sure if media streams expose enough control for the UA to be able to show controls for streams where it's possible to have them.

cabanier:
> I haven't looked really closely at this yet (will tomorrow), but have we considered making this based on media streams? This means videos can still be piped to WebXR, but so can webrtc streams and getUserMedia.

It seems the media streams are indeed more flexible but won't the video element take care of thing like cross-origin and the actual playing of the stream? It seems we would have to add all that logic to the media layers which might be hard to get right.

> That said, when it's a video element layer the UA can actually choose to show video controls. I'm not sure if media streams expose enough control for the UA to be able to show controls for streams where it's possible to have them.

The spec currently defines that there are no media controls shown. We won't be able to do this properly until we have DOM layers.

Manishearth:
> but won't the video element take care of things like cross-origin and the actual playing of the stream? 

Right, when you want a URL-based stream you still create a video element, you just [extract its stream](https://w3c.github.io/mediacapture-fromelement/#dom-htmlmediaelement-capturestream). Streams themselves don't have an origin, if you have access to one you can play it.

That said, video elements might actually be the more flexible option. You can also attach media streams to video elements, so if you want to stream webrtc you can attach it to a video element and then stick the video element in a layer. This preserves the ability to play/pause/seek video content that supports it since `<video src>` supports that, even if `MediaStream` does not.

I have a feeling that such APIs are _supposed_ to use media streams from an architectural POV, though. But if they don't support play/pause/seek we're kinda forced to use `<video>`.

> The spec currently defines that there are no media controls shown. We won't be able to do this properly until we have DOM layers.

We can still allow UAs to display controls, no? Implementation-wise, sure, you may need some level of DOM overlay support to do this nicely, but UAs can composite a play/pause button on hover. Hubs, for example, does this in software.

That said there is no concept of "hover" or "click" on a layer until we define the pointer XRRay from https://github.com/immersive-web/layers/issues/21. (that keeps cropping up, maybe I should write some spec text for that!)

toji:
This is interesting, as I wasn't super familiar with MediaStreams previously. It's worth considering. We've already heard some feedback (from @thetuvix, for example) that using the `video` element directly may overtly tie us to the DOM and imply support for rendering media controls in cases where we don't intend to support it.

> when you want a URL-based stream you still create a video element, you just extract its stream.
> ...This preserves the ability to play/pause/seek video content that supports it since <video src> supports that, even if MediaStream does not.

If you extract a `MediaStream` from a `video` element, does it's playback remain connected to the `video` element itself? If so then I don't see an inherent advantage to using `video` elements over `MediaStream`s, but I'd also like to see a deeper analysis done to determine if there's any limitations that would make it inappropriate for use by WebXR. For example, there's a some text in the `captureStream()` spec that indicates that when the video is stopped and restarted it produces new `MediaStreamTrack`s.

> Since a MediaStreamTrack can only end once, a track that is enabled, disabled and re-enabled will be captured as two separate tracks. Similarly, restarting playback after playback ends causes a new set of captured MediaStreamTrack instances to be created. Seeking during playback without changing track selection does not generate events or cause a captured MediaStreamTrack to end.

I think these get attached to the original `MediaStream`, and thus the XR layer could continue displaying it without intervention, but I'm not 100% sure.

cabanier:
Do you want me to file an issue about switching from <video> to media streams?
Then we can merge this PR (which doesn't deal with the actual transfer of the video  pixels yet) and continue the discussion in the issue.

Manishearth:
> If you extract a MediaStream from a video element, does it's playback remain connected to the video element itself?

Yes. But spec-wise if we wanted to specify "if the stream comes from a video element you may show controls" I'm not sure if it's easy to.

> that using the video element directly may overtly tie us to the DOM and imply support for rendering media controls in cases where we don't intend to support it.

hmm, good point

another option is to accept `MediaStream or HTMLVideoElement`, where media streams are live streams that the layer has no control over, and HTMLVideoElements may show some UA controls over the layer.

>  Then we can merge this PR (which doesn't deal with the actual transfer of the video pixels yet) and continue the discussion in the issue.

Sounds good, I'll give you a quick review here

webxrlayers-1.bs
```diff
@@ -291,7 +293,8 @@ When calling {{destroy()}} on {{XRCompositionLayer}} |layer|, the user agent MUS
 </div>
 
 Each {{XRCompositionLayer}} has a <dfn for="XRCompositionLayer">context</dfn> object which is an instance
-of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.
+of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}} and a <dfn for="XRCompositionLayer">media</dfn>
+object which is an instance of a {{HTMLVideoElement}}.
```
Manishearth, line 297: make sure to declare it nullable?

webxrlayers-1.bs
```diff
@@ -1034,6 +1056,116 @@ interface XRMediaBinding {
...
+{{XRSession}} it was created with.
+
+NOTE: It is possible to create more than one {{XRMediaBinding}}. The lifetime of a layer is not tied
+to the lifetime of the {{XRMediaBinding}} that created it.
+
+Each layer created through {{XRMediaBinding}} has an internal {{HTMLVideoElement}} [=XRCompositionLayer/media=].
+If the layer is part of the session's {{XRSession/renderState}}, it will display the current frame of the video. The
+layer is update at the native framerate of the [=/XR device=] or the video, whichever is less.
+
+NOTE: only the video frames will be displayed in the layer. Video controls should be implemented by the author and must
```
Manishearth, line 1069: unsure if we should have this, but it can be discussed in another issue

Manishearth: maybe have an ISSUE line for this

cabanier: done

PR #116, opened 2020-04-30
changed hardcoded number to number of views

### LINKED ISSUE #114: How should layers behave when there's more than two views of the scene?
For example, devices may have a third view for an on-board camera, used when generating screen recordings (so the XR content lines up with the real-world content).

cabanier:
fixes #114

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -656,13 +656,14 @@ To <dfn>allocate color textures</dfn> using a {{XRLayerInit/layout}} |layout|, a
         <dt> Otherwise
           <dd> Initialize |array| with 1 new instance of {{WebGLTexture}} created as a {{TEXTURE_2D}} texture with |context| and |init|'s {{XRLayerInit/alpha}}, {{XRLayerInit/viewPixelWidth}} and {{XRLayerInit/viewPixelHeight}} values.
         <dl>
+  1. let |numViews| be the number of all the session's {{XRView|XRViews}}.
   1. If |layout| is {{XRLayerLayout/stereo}}:
         <dl class="switch">
         <dt> If |textureType| is {{"texture-array"}}:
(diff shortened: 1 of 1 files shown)


cabanier:
> There are a couple of uses of the word "double" left at https://github.com/cabanier/layers/blob/textureperview/webxrlayers-1.bs#L669-L670 which should become "numViews times". Apart from that, LGTM, thanks!

Those need to stay. Those layouts are designed for stereo artwork (such as video) and it will be unexpected that they suddenly need an extra view.

asajeffrey:
> > There are a couple of uses of the word "double" left at https://github.com/cabanier/layers/blob/textureperview/webxrlayers-1.bs#L669-L670 which should become "numViews times". Apart from that, LGTM, thanks!
> 
> Those need to stay. Those layouts are designed for stereo artwork (such as video) and it will be unexpected that they suddenly need an extra view.

Hmm, in that case should `stereo-left-right` and `stereo-top-bottom` throw an exception if there's anything other than two views?

cabanier:
> > Those need to stay. Those layouts are designed for stereo artwork (such as video) and it will be unexpected that they suddenly need an extra view.
> 
> Hmm, in that case should `stereo-left-right` and `stereo-top-bottom` throw an exception if there's anything other than two views?

Would having more than 2 views be feature that the author is opting in?
If so, maybe it's ok to have more than 2 views for a side-by-side texture.

asajeffrey:
Hmm, interesting question about how the number of views would be negotiated. It would be nice if there common case was that content would work on all devices, but there are odd cases like spectator view, where it's not obvious how something like stereo video should be composited if there's a spectator view.

cabanier:
> Hmm, interesting question about how the number of views would be negotiated. It would be nice if there common case was that content would work on all devices, but there are odd cases like spectator view, where it's not obvious how something like stereo video should be composited if there's a spectator view.

Yes, we probably would display the left or right video.
Let's merge the PR for now and see if we can merge a new view name (`camera`?) into WebXR and then we can special case it in the layers spec.

asajeffrey:
Sound good, thanks for making the changes!

PR #121, opened 2020-05-06
First pass on redraw event

cabanier:


webxrlayers-1.bs
```diff
@@ -1165,6 +1191,36 @@ When this method is invoked, the user agent MUST run the following steps:
...
+XRLayerEvent {#xrlayerevent-interface}
+--------------
+
+{{XRLayerEvent}} is fired to indicate changes to the state of an {{XRLayer}}.
+
+<pre class="idl">
+[SecureContext, Exposed=Window]
+interface XRLayerEvent : Event {
+  constructor(DOMString type, XRLayerEventInit eventInitDict);
+  [SameObject] readonly attribute XRLayer session;
```
toji, line 1206:
```suggestion
  [SameObject] readonly attribute XRLayer layer;
```

I assume this was the intent?

webxrlayers-1.bs
```diff
@@ -275,11 +275,17 @@ enable optical chromatic aberration correction for the layer. Only valid if this
...
+might cause flickering or other side effects.
 
-NOTE: the user agent might set this flag if the underlying resources were reallocated or if the [=XR Compositor=] can no
-longer reproject the layer. Failing to redraw the content in the next [=XR animation frame=] might cause flickering or
-other side effects.
+<div class="algorithm" data-algorithm="redrawLayerAlgo">
+When <dfn>the underlying resources of a layer are reallocated</dfn> for an {{XRCompositionLayer}} |layer|,
+the user agent MUST run the following steps:
+  1. Set the |layer|'s {{XRCompositionLayer/needsRedraw}} array to an empty array.
+  1. If |layer| is not a {{XRProjectionLayer}}, [=queue a task=] to [=fire an event=] named {{redraw}} on |layer|.
```
toji, line 286: I like the omission of `onredraw` from the projection layer specifically, because we really just want devs to redraw that all the time anyway. 👍

webxrlayers-1.bs
cabanier: yes. Thank you!

PR #122, opened 2020-05-06
minimize new instances of XRWebGLSubImage

cabanier:


webxrlayers-1.bs
```diff
@@ -967,24 +968,24 @@ To <dfn>validate the state of the XRWebGLSubImage creation function</dfn> of a {
 <div class="algorithm" data-algorithm="getSubImageAlgo">
 The <dfn method for="XRWebGLBinding">getSubImage(XRCompositionLayer |layer|, XRFrame |frame|)</dfn> method creates a new {{XRWebGLSubImage}}.
 
-When this method is invoked, it MUST run the following steps:
+When this method is invoked on a {{XRWebGLBinding}} |binding|, it MUST run the following steps:
```
toji, line 971: I don't know that this works. It seems like a valid pattern to query all the SubImages at the top of the render loop and then switch between them on a per-object basis to minimize state change (The WebXR samples do this in most of the later samples). With this algorithm, though, you'd inadvertently end up with N references to the last view's subImage.

I do think it should be possible for each layer to contain one pre-allocated sub image per view or stereo pair and return that with a new set of textures each frame. That at least caps the sub image allocation to `NumViews * NumLayers`

webxrlayers-1.bs
```diff
@@ -532,8 +532,8 @@ The {{XRWebGLSubImage}} object is used during rendering of the layer.
 
 <pre class="idl">
 interface XRWebGLSubImage : XRSubImage {
-  readonly attribute WebGLTexture colorTexture;
-  readonly attribute WebGLTexture? depthStencilTexture;
+  [SameObject] readonly attribute WebGLTexture colorTexture;
```
toji, line 535: If we are re-using sub images in any capacity I think think this works, since the WebGLTexture will change frame-to-frame. Unless we're planning on having the underlying native texture swapped out while the JS wrapper stays the same, but that sounds questionable.

cabanier: > Unless we're planning on having the underlying native texture swapped out while the JS wrapper stays the same, but that sounds questionable.

yes, that will never happen:-)

webxrlayers-1.bs
cabanier: ok. I think that means that the cached binding should live in the layer. 
I'll make the necessary changes.

cabanier: hmm, that doesn't work well because then media layer and future webgpu layers have useless instances of XRWebGLSubImage.
It seems the cache needs to live inside XRWebGLBinding.

toji: That's probably fine, it'd just need to be stored in some sort of Layer->SubImage[] map.

I'll point out that in Chrome I would be inclined to implement this as having internal subclasses for each layer type that host the WebGL resource. Such as:

```cpp
class XRWebGLProjectionLayer : public XRProjectionLayer {
  //...
  HashMap<XRView*, XRWebGLSubImage*> subImages;
}```

Comment (cabanier) [reply webxrlayers-1.bs:971]: When looking at [getElementsByTagName](https://dom.spec.whatwg.org/#concept-getelementsbytagnamens), the HTML spec just states:

> When invoked with the same arguments, the same HTMLCollection object may be returned as returned by an earlier call.

I wonder if we can do the something similar here.
For instance:

1.  Initialize subimage as follows:
   if getViewSubImage was called on the same binding, layer and view, the user agent MAY:
       let subimage be the same object as returned by an earlier call with the same arguments
  else
       let submimage be a new subimage

**Comment by Manishearth** [comment]
deferring to brandon's review

File: webxrlayers-1.bs [hunk 3]
```diff
@@ -960,31 +960,43 @@ To <dfn>validate the state of the XRWebGLSubImage creation function</dfn> of a {
...
 <div class="algorithm" data-algorithm="getSubImageAlgo">
 The <dfn method for="XRWebGLBinding">getSubImage(XRCompositionLayer |layer|, XRFrame |frame|)</dfn> method creates a new {{XRWebGLSubImage}}.
 
-When this method is invoked, it MUST run the following steps:
+When this method is invoked on a {{XRWebGLBinding}} |binding|, it MUST run the following steps:
 
-  1. Let |subimage| be a new {{XRWebGLSubImage}}
+  1. Initialize |subimage| as follows:
+    <dl class="switch">
+      <dt> If {{XRWebGLBinding/getSubImage()}} was called previously with the same |binding|, |layer| and |frame|, the user agent MAY:
```
toji, line 974: This feels much better! Thanks!

toji:
LGTM. Merged.

PR #124, opened 2020-05-07
 moved viewPixelWidth/Height to XRWebGLSubImage

### LINKED ISSUE #123: viewPixelWidth and viewPixelHeight semantics?
https://immersive-web.github.io/layers/#dom-xrcompositionlayer-viewpixelwidth says that "The viewPixelWidth and viewPixelHeight attributes return the width and height of the GL attachments, respectively. If there are no attachments, they will return 0."

What should these return if called outside an rAF? (There may or may not be GL attachments outside of an rAF, depending on the implementation).

For `stereo-left-right` and `stereo-top-bottom` are these the dimensions of one eye, or the dimensions of the whole texture?

cabanier:
Closes #123

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -601,7 +601,7 @@ The <a href="https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pd
 section in the OpenGL ES 3.0 spec defines these limitations in more detail.
 
 The <dfn attribute for="XRWebGLSubImage">textureWidth</dfn> and <dfn attribute for="XRWebGLSubImage">textureHeight</dfn> attributes
-return the width and height of the GL attachments, respectively.
+return the width and height in pixels of the GL attachments, respectively.
 
 XRTextureType {#xrtexturetype}
 -------------
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -574,8 +574,8 @@ interface XRWebGLSubImage : XRSubImage {
   [SameObject] readonly attribute WebGLTexture colorTexture;
   [SameObject] readonly attribute WebGLTexture? depthStencilTexture;
   readonly attribute unsigned long? imageIndex;
-  readonly attribute unsigned long viewPixelWidth;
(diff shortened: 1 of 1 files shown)


toji:
If you've got the sub image you can figure out the view size by by looking at the viewport width/height. I think we should expose the texture width and height, though, because it's a bit of a pain to pull from WebGL consistently and can be necessary for certain effects.

The previous conversation about the rename to viewPixelWidth/viewPixelHeight was (if I recall correctly) primarily about what the dictionary inputs should be called. It makes sense for the user to specify things in view size, but the later to report the full texture size.

cabanier:
> The previous conversation about the rename to viewPixelWidth/viewPixelHeight was (if I recall correctly) primarily about what the dictionary inputs should be called. It makes sense for the user to specify things in view size, but the later to report the full texture size.

That sounds reasonable. Should I rename the attributes to `pixelWidth`/`pixelHeight` then?

toji:
I feel like maybe `textureWidth/Height` is less ambiguous as to what it's referring to? I think it's well understood that texture sizes are in pixels, and if it just says `pixelWidth/Height` devs might assume it's referring to the view's pixel size.

webxrlayers-1.bs
```diff
@@ -603,6 +600,8 @@ being generated, even if it does not affect the dimensions or format: TexImage*,
 The <a href="https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf#nameddest=section-3.8.4">"Immutable-Format Texture Images"</a>
 section in the OpenGL ES 3.0 spec defines these limitations in more detail.
 
+The <dfn attribute for="XRWebGLSubImage">textureWidth</dfn> and <dfn attribute for="XRWebGLSubImage">textureHeight</dfn> attributes
+return the width and height of the GL attachments, respectively.
```
asajeffrey, line 604: Might be worth saying that they're measured in pixels?

Manishearth:
deferring to alan and brandon's review on this one, since it's mostly GL stuff

Issue #112, opened 2020-04-29
Add fixed foveation optional attribute to ProjectionLayer


Artyom17:
Some HW allows to use so called 'fixed foveation rendering' (FFR) techniques, applicable only to projection layer. Basically, the idea behind it is to render less on borders (vision periphery) than in the center. It is handled by the GL driver, it is pretty much free and requires no modifications to shaders (unlike multiview) and it saves GPU perf (while multiview saves CPU). GL driver just rasterizes with less resolution at periphery (on tile basis for mobile). 

There could be different levels of FFR applied, the higher level is the more visible artifacts are but the more perf is gained. Depending on content's complexity the win could be up to 25% of GPU time.

We can specify it by float 0.0 ... 1.0f, where 0.0 indicates no foveation and 1.0 - max possible level. For example in Oculus we currently have 4 levels: no foveation, low, mid, high, which will translate to 0.0, 0.3, 0.6, 1.0. If something like 0.5 specified then UA can decide which actual level to select (but this will be vendor and HW-specific, I assume).

As I said, it makes sense only for ProjectionLayers.

cabanier:
Is this something specific to Oculus or is it generally available?
Also, do you have a link to a description of this property?

Artyom17:
It is not Oculus-specific, but it is not very widely available either. Basically, this feature is based on QCOM GL extension (and Vulkan version is available too) - https://www.khronos.org/registry/OpenGL/extensions/QCOM/QCOM_texture_foveated.txt

Artyom17:
Here is the link to description of this extension for WebVR: https://developer.oculus.com/documentation/oculus-browser/browser-optimize-rendering/

cabanier:
> It is not Oculus-specific, but it is not very widely available either. Basically, this feature is based on QCOM GL extension (and Vulkan version is available too) - https://www.khronos.org/registry/OpenGL/extensions/QCOM/QCOM_texture_foveated.txt

It seems there are a lot of potential flags that are not always applicable to all devices (ie does a phone need `chromaticAberrationCorrection`?)
Instead of adding various attributes, maybe we can provide some sort of mechanism to let the author know what features are available and then let them control them.

It could look like this:
```webidl
partial interface XRCompositionLayer {
  readonly attribute FrozenArray<DOMString> extensions;
  any getExtension(DOMString);
  void setExtension(DOMString, any);
};
```

cabanier:
[...]

PR #134, opened 2020-05-08
strawman proposal for layer parameters

cabanier:
This is the first proposal to address #112.
Basically, the parameter will exists on CompositionLayer if the layer supports it.

The dictionary is passed during construction of the layers or by calling `setParameters` at runtime. 
`setParameters` can be called at any time but won't take effect until the next rAF.

What do people think?

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -266,20 +266,51 @@ interface XRCompositionLayer : XRLayer {
 
 The <dfn attribute for="XRCompositionLayer">layout</dfn> attribute returns the layout of the layer.
 
-The <dfn attribute for="XRCompositionLayer">blendTextureSourceAlpha</dfn> attribute enables the layer’s texture alpha channel. It is defined if the UA supports this parameter.
+The <dfn attribute for="XRCompositionLayer">blendTextureSourceAlpha</dfn> attribute reflects if the layer has a texture alpha channel.
+It is <code>null</code> if the UA does not support this parameter.
 
 The <dfn attribute for="XRCompositionLayer">chromaticAberrationCorrection</dfn> attribute is a hint for the [=XR Compositor=] to
-enable optical chromatic aberration correction for the layer. It is defined if the UA supports this parameter.
+enable optical chromatic aberration correction for the layer.
+It is <code>null</code> if the UA does not support this parameter.
 
 ISSUE: describe what chromaticAberrationCorrection does.
 
-The <dfn attribute for="XRCompositionLayer">fixedFoveation</dfn> controls the amount of foveation. It is defined if the UA supports this parameter.
+The <dfn attribute for="XRCompositionLayer">fixedFoveation</dfn> controls the amount of foveation.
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -262,13 +266,21 @@ interface XRCompositionLayer : XRLayer {
 
 The <dfn attribute for="XRCompositionLayer">layout</dfn> attribute returns the layout of the layer.
 
-The <dfn attribute for="XRCompositionLayer">blendTextureSourceAlpha</dfn> attribute enables the layer’s texture alpha channel.
+The <dfn attribute for="XRCompositionLayer">blendTextureSourceAlpha</dfn> attribute enables the layer’s texture alpha channel. It is defined if the UA supports this parameter.
```
Manishearth, line 269: It's a readonly attribute, it doesn't enable anything.

```suggestion
The <dfn attribute for="XRCompositionLayer">blendTextureSourceAlpha</dfn> attribute reflects if the layer has a texture alpha channel. It <code>null</code> if the UA does not support this parameter..
```

cabanier:
> not really a fan of the "single method that tweaks multiple settings" pattern, it gets a bit hairy and it's easy to accidentally reset things if you're not paying attention.

True. Is there a way to set the dictionary so only the parameters that the author inserted get used?
Maybe I just need to remove the defaults and define behavior in prose.

> feels like we should either have individual setter methods or use r/w attributes that throw.

One of the issue I tried to address is that you can't change the parameter during a rAF. I worked around by having a method that queues a task. This still allows changes but that will only be reflected after the rAF ends.
If it's just an attribute, it would be weird that the value doesn't change right away.

Manishearth:
> True. Is there a way to set the dictionary so only the parameters that the author inserted get used?

remove the defaults, make everything optional (might have to make things nullable, i forget)

>  If it's just an attribute, it would be weird that the value doesn't change right away.

okay, in that case `setFoo` methods might work better

> One of the issue I tried to address is that you can't change the parameter during a rAF

This feels like it's similar to updateRenderState, which has some pretty fiddly semantics. But i think having separate setters would make this easier

cabanier:
> > True. Is there a way to set the dictionary so only the parameters that the author inserted get used?
> 
> remove the defaults, make everything optional (might have to make things nullable, i forget)

ok. I did that in the latest PR (but I still need to create a default initialisation)

> > If it's just an attribute, it would be weird that the value doesn't change right away.
> 
> okay, in that case `setFoo` methods might work better
> 
> > One of the issue I tried to address is that you can't change the parameter during a rAF
> 
> This feels like it's similar to updateRenderState, which has some pretty fiddly semantics. But i think having separate setters would make this easier

I agree it's fiddly, but the setters suffer from the same problem. I think I prefer a single set call that reuse the same dictionary that is used during initialization.

cabanier:
/agenda should we move back to regular attributes and specify how their state is reflected?

cabanier:
During this week's meeting, this was discussed and we decided that regular attributes that define when they apply is easier.
Closing this PR.

Issue #46, opened 2020-04-04
Add 'layout' hint for quad, cylinder, equirect stereo layers


Artyom17:
Displaying pre-rendered stereo images or videos will require specification of stereo layout, which is currently mentioned for video quad layer. However, for stereo photos, stereo 180/360 panoramas it is also required WHEN texture2D is used. For texture arrays it will be ignored.

cabanier:
@Artyom17 and I talked offline about this.
His concern was that for certain sizes, the UA won't be able to support a left to right texture because it exceeds the maximum texture size.

However, the current API allows the UA to switch to a different layout or even another texture if the requested size is too big for a left-to-right texture. At best, if the author knows if it's left-to-right or top-to-bottom, they could blit for 2 eyes at the same time.
For now, we don't need the hint but we might revisit this later.

Artyom17:
>> For now, we don't need the hint
I disagree. If I use pre-rendered content, such as stereo 360 photos, or regular 3D photos, I want dev to be able to use the texture unmodified. I don't understand why we are going to limit to side-by-side stereo only and not to include top-to-bottom as well. I'd strongly recommend to add the hint to all the layers (well, other than projection layer).

cabanier:
The only thing this would solve is a blit during rendering (so 2 instead of 1).
It also forces the UA to support all layout types. Currently they can optimize and only offer what is efficient.

If we offer a hint, do we also extend the API to say what type of layout was created by the UA? Can the hint be ignored?

Artyom17:
What I think:

* Each blit takes at least 0.5 ms on Quest, those are not cheap at all on mobile (of course depends on size and stereo 360 content is usually of big resolution). Especially when you update it every frame (more applicable to video)
* We already have those layouts for video, so supporting layouts for other layers is not a problem at all.
* Hint can't be ignored, otherwise it will defeat its purpose. Projection layer can be different. 
* layout can be returned by layer as readonly attribute
I even think we don't need a 'boolean stereo', we need an enum XRStereoLayout instead, with members like:
"mono" - mono
"stereo" - default stereo, UA is selecting which layout is used; also used for texture arrays;
"stereo-left-to-right"
"stereo-top-bottom"

Artyom17:
I am currently prototyping stereo support for Equirect (and same will be applicable to Cylinder and Quad, while Cube might be different). What I also found out:
[...]

PR #140, opened 2020-05-16
Add XRCompositor language to accomodate monoscopic displays

### LINKED ISSUE #68: How should layers behave on devices that have only 1 screen
For instance, projection layers are currently defined to always be `stereo` but that is not the case on mobile phones

cabanier:
+ clarified stereo 
+ fixed some typos

Closes #68


Manishearth:
Hmm. This might be confusing, it might be worth erroring if these are used instead of mono?

Alternatively, we can perhaps rename these to not be "stereo" but instead be "views" where they each have the number of subrects equal to the number of views, and we can have `getViewport()`-esque stuff to get viewports instead of requiring right-left and top-bottom. It seems very weird to me that the core spec is super flexible on this but layers hardcodes the choices.

This would also properly solve https://github.com/immersive-web/layers/issues/114

cabanier:
> Hmm. This might be confusing, it might be worth erroring if these are used instead of mono?

No, we don't want to error out because then mono experiences will crash unless the author specifically accommodates them.

> Alternatively, we can perhaps rename these to not be "stereo" but instead be "views" where they each have the number of subrects equal to the number of views, and we can have `getViewport()`-esque stuff to get viewports instead of requiring right-left and top-bottom. It seems very weird to me that the core spec is super flexible on this but layers hardcodes the choices.

I agree that `stereo` doesn't feel right but `views` is also confusing.
Maybe `Natural`, `all` or `default`?

Manishearth:
yeah I don't quite care about the naming :)

It seems like we should just have "mono" vs "natural" and natural is laid out however the UA wants, using something like getViewport(). If we do this are the horizontal and vertical variants necessary?  I.e. does the user still need the ability to request a particular arrangement?

cabanier:
> yeah I don't quite care about the naming :)
> 
> It seems like we should just have "mono" vs "natural" and natural is laid out however the UA wants, using something like getViewport().

Should I change it to `natural` or should we bring it to the group?

> If we do this are the horizontal and vertical variants necessary? I.e. does the user still need the ability to request a particular arrangement?

Yes, those are still needed to accommodate pre-canned video, bitmaps, etc

cabanier:
/agenda What should we replace the `stereo` layout with?

Manishearth:
> Yes, those are still needed to accommodate pre-canned video, bitmaps, etc

Should this not just use the XRMediaBinding stuff?

Even if we decide to have such layers, it still feels weird to then patch mono support in for these layers.

Overall these two seem to b pretty specal-case-y and I'm wary of adding something like that.

cabanier:
> > Yes, those are still needed to accommodate pre-canned video, bitmaps, etc
> 
> Should this not just use the XRMediaBinding stuff?

No because authors will want to add other effects on top of video or use the stereo bitmaps.

> Even if we decide to have such layers, it still feels weird to then patch mono support in for these layers.
> 
> Overall these two seem to b pretty specal-case-y and I'm wary of adding something like that.

I'm not adding side-by-side layout in the PR. That was already discussed in previous issues.

Manishearth:
> I'm not adding side-by-side layout in the PR. That was already discussed in previous issues.

Yes, but you are deciding a behavior for side-by-side layout textures in mono sessions, which is what's weird here.

And yeah I think this is a wider discussion that needs to be had: it was kinda ongoing in #114 but that was closed with a specific fix. I'll open something new.

cabanier:
> > I'm not adding side-by-side layout in the PR. That was already discussed in previous issues.
> 
> Yes, but you are deciding a behavior for side-by-side layout textures in mono sessions, which is what's weird here.

If you have other behavior in mind, we can discuss that here.
Side-by-side layout should not break workflows that happen to be monoscopic.

Manishearth:
>  Side-by-side layout should not break workflows that happen to be monoscopic.

Hmm. It still feels weird, but I won't block this. I'll file a separate issue for the existence of left-right and top-bottom.

cabanier:
> > Side-by-side layout should not break workflows that happen to be monoscopic.
> 
> Hmm. It still feels weird, but I won't block this. I'll file a separate issue for the existence of left-right and top-bottom.

It was introduced by #46. We also discussed it in a weekly call but I don't know how to link that.

Manishearth:
Yeah, I think there's a bit more to it than that. For example, even if we rename "stereo" there still needs to be a way to get viewports.

Manishearth:
Hmm, looking at it it seems fine now, but we should probably add non-normative notes explaining these modes better.

cabanier:
> Hmm, looking at it it seems fine now, but we should probably add non-normative notes explaining these modes better.

@Manishearth I added a note. Do you think more information is needed?

Manishearth:
Looks great, thanks!

cabanier:
We discussed this change a bit in this week's call but we didn't resolve what name we should use.
I'm leaning towards `default`

Also, @toji I was wrong when I said a projection layers can be mono on a stereo device. It's defined to be always `stereo` (aka `default`) so it has the same number of subimages as the session.

cabanier:
@toji any comments on my latest changes?

webxrlayers-1.bs
```diff
@@ -240,18 +240,22 @@ The {{XRLayerLayout}} enum defines the layout of the layer.
 <pre class="idl">
 enum XRLayerLayout {
   "mono",
-  "stereo",
+  "default",
```
toji, line 243: This feels fine to me, although as the most minor nit ever: If we're gonna call it "default" then we should probably make it the first item in the enum. :)

cabanier: ok :-)

PR #145, opened 2020-05-20
Added fixed foveation + added language about optional attributes

### LINKED ISSUE #112: Add fixed foveation optional attribute to ProjectionLayer

cabanier:
Closes #112

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -262,7 +262,9 @@ interface XRCompositionLayer : XRLayer {
   readonly attribute XRLayerLayout layout;
 
   attribute boolean blendTextureSourceAlpha;
-  attribute boolean chromaticAberrationCorrection;
+  attribute boolean? chromaticAberrationCorrection;
+  attribute float? fixedFoveation;
+
   readonly attribute boolean needsRedraw;
 
   void destroy();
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -274,9 +276,13 @@ The <dfn attribute for="XRCompositionLayer">layout</dfn> attribute returns the l
 The <dfn attribute for="XRCompositionLayer">blendTextureSourceAlpha</dfn> attribute enables the layer’s texture alpha channel.
 
 The <dfn attribute for="XRCompositionLayer">chromaticAberrationCorrection</dfn> attribute is a hint for the [=XR Compositor=] to
-enable optical chromatic aberration correction for the layer. Only valid if this isn't done by default.
+enable optical chromatic aberration correction for the layer. This attribute is only defined if it is supported by the user agent.
 
-ISSUE: describe what chromaticAberrationCorrection does.
+The <dfn attribute for="XRCompositionLayer">fixedFoveation</dfn> attribute controls the amount of foveation used by the
+[=XR Compositor=]. This attribute is only defined if the user agent supports it.
```
Manishearth, line 282: "This attribute is only defined if the user agent supports it.": this isn't kosher.

We should instead say: If the user agent or device does not support this attribute, they should return `null` on getting.

Manishearth: Ideally you write a two-line algorithm for each of setting and getting but that's not a huge deal.

cabanier: > Ideally you write a two-line algorithm for each of setting and getting but that's not a huge deal.

The algorithm for initializing the attributes takes care of that.

cabanier: > "This attribute is only defined if the user agent supports it.": this isn't kosher.
> 
> We should instead say: If the user agent or device does not support this attribute, they should return `null` on getting, and setting should be a no-op.

Maybe: If the user agent or device does not support this attribute, this attribute is undefined.

Manishearth: The algorithm doesn't take care of that, you can't define a webidl attribute as being conditionally defined, you can only define it's setters to return null.

If the user agent supports this module it is a spec requirement to support the attribute as an attribute. We should give them the power to force it to be always-null, but we cannot spec it as undefined.

Aside: are there cases where the user agent supports this attribute but the device does not? Perhaps the nullability should be per device.

Manishearth: I believe it's actually technically incorrect to use an attribute as both a slot and something with get/set semantics as well, but it's generally understood what it means when you do that so that's not as big a deal (we do this in a bunch of places in WebXR too). Some specs go all out and define an internal member for each attribute that reflects internal state.

cabanier: ok. I assumed '?' meant `undefined` and not `null`. I will make changes to the PR.

PR #149, opened 2020-05-28
Remove optional from layers attribute

cabanier:
the layers array will always be populated because projection layers are supported by default


toji:
What's the intended behavior if the developer sets baseLayer instead?

toji:
> There's no reason for an array to be both nullable and empty-able

Post-merge +1 to that.

PR #150, opened 2020-05-30
removed w3c.json since this repository is not yet moved to WG from CG

himorin:
@dontcallmedom , do we need to have w3c.json with not rec-track repo-type? (sorry, just forgot and/or not following direction...)

### DIFF
--- a/w3c.json
+++ b/w3c.json
@@ -1,5 +1,0 @@
- {
-    "group":      [109735]
-,   "contacts":   ["cabanier"]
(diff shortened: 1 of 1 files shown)


dontcallmedom:
@himorin having a w3c.json for non-rec track repo (with "repo-type" set to "cg-report") is useful (but not mandatory)

himorin:
(noted, I'll do it from next time)

PR #157, opened 2020-06-08
Explained XRCompositor behavior for additional displays

### LINKED ISSUE #155: Presenting stereo layers to devices with more than two views
https://immersive-web.github.io/layers/#xrcompositorchanges discusses how to present stereo layers when there's only one view, but not when there's more than two (e.g. devices which record to a third view).

cabanier:
Closes #155

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -1446,6 +1446,10 @@ If the [=XR Compositor=] is driving a monoscopic display and the rendered {{XRCo
 If the {{XRCompositionLayer}} was created through {{XRMediaBinding}} and the {{XRMediaLayerInit/"invertStereo"}} flag was set to
 <code>true</code>, the bottom portion of the {{XRLayer}} is shown while the top portion is ignored.
 
+If the [=XR Compositor=] is driving a additional display in addition to a monoscopic or stereoscopic one, it is up to the
+user agent how a {{XRCompositionLayer}} with an {{XRLayerLayout}} of {{XRLayerLayout/"stereo-left-right"}} or
+{{XRLayerLayout/"stereo-top-bottom"}} is shown to that additional display.
+
 Animation frames changes {#animationframeschanges}
 ------------------------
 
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -1446,6 +1446,10 @@ If the [=XR Compositor=] is driving a monoscopic display and the rendered {{XRCo
 If the {{XRCompositionLayer}} was created through {{XRMediaBinding}} and the {{XRMediaLayerInit/"invertStereo"}} flag was set to
 <code>true</code>, the bottom portion of the {{XRLayer}} is shown while the top portion is ignored.
 
+If the [=XR Compositor=] is driving a additional display in addition to a monoscopic or stereoscopic one, it is up to the
```
asajeffrey, line 1449: Possibly more than one additional displays?

webxrlayers-1.bs
```diff
@@ -1446,6 +1446,10 @@ If the [=XR Compositor=] is driving a monoscopic display and the rendered {{XRCo
...
+user agent how a {{XRCompositionLayer}} with an {{XRLayerLayout}} of {{XRLayerLayout/"stereo-left-right"}} or
+{{XRLayerLayout/"stereo-top-bottom"}} is shown to that additional display.
```
asajeffrey, line 1451: Do we need language about `mono` too?

webxrlayers-1.bs
cabanier: Fixed with latest PR.

webxrlayers-1.bs
cabanier: Isn't that already covered by the 2 previous paragraphs?

asajeffrey: is it, I don't see any text about how to render mono layers. I mean it's kind of obvious, so maybe it doesn't need spelled out?

cabanier: Yes, I don't think it's needed.

PR #160, opened 2020-06-10
Fix for allocating textures

cabanier:
After the latest changes, the spec was incorrect for non-projection layers. 
This change breaks projection layers into their own algorithm and introduces a recommended resolution for views.

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -255,7 +255,7 @@ enum XRLayerLayout {
 };
 </pre>
 
- - A layout of <dfn enum-value for="XRLayerLayout">default</dfn> indicates that the layer can accomodate all the session’s {{XRView|XRViews}}.
+ - A layout of <dfn enum-value for="XRLayerLayout">default</dfn> indicates that the layer can accomodate all the views of session’s [=list of views=].
  - A layout of <dfn enum-value for="XRLayerLayout">mono</dfn> indicates that the layer is mono.
  - A layout of <dfn enum-value for="XRLayerLayout">stereo-left-right</dfn> indicates that the layer is divided left to right.
  - A layout of <dfn enum-value for="XRLayerLayout">stereo-top-bottom</dfn> indicates that the layer is divided top to bottom.
@@ -1170,7 +1170,7 @@ When this method is invoked, the user agent MUST run the following steps:
   1. Initialize |layer|'s {{XRCompositionLayer/needsRedraw}} to <code>true</code>.
   1. If |layout| is {{XRLayerLayout/"stereo-left-right"}} or {{XRLayerLayout/"stereo-top-bottom"}}, throw {{TypeError}} and abort these steps.
   1. Let |layer|'s [=colorTextures=] be a [=new=] array in the [=relevant realm=] of this {{XRCubeLayer}}.
-  1. let |numViews| be the number of all the session's {{XRView|XRViews}}.
+  1. let |numViews| be the number of |session|'s [=list of views=].
   1. Initialize |layer|'s [=colorTextures=] as follows, based on the value of |layout|:
     <dl class="switch">
       <dt> {{XRLayerLayout/"mono"}}:
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -938,6 +938,7 @@ To <dfn>allocate color textures for projection layers</dfn> using an {{XRProject
             1. Append |texture| to |array|.
           <dd> Return |array| and abort these steps.
         </dl>
+  1. If the session’s [=view|views=] in the [=list of views=] don't all have the same [=recommended WebGL texture resolution=], throw an {{NotSupportedError}} and abort these steps.
   1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/stereo-left-right}}, initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture using |context| , |alpha|, double of |width| and |height|.
   1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/stereo-top-bottom}}, initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture using |context| , |alpha|, |width| and double of |height|.
   1. return |array|.
@@ -970,6 +971,7 @@ To <dfn>allocate depth textures for projection layers</dfn> using an {{XRProject
             1. Append |texture| to |array|.
           <dd> Return |array| and abort these steps.
         </dl>
+  1. If the session’s [=view|views=] in the [=list of views=] don't all have the same [=recommended WebGL texture resolution=], throw an {{NotSupportedError}} and abort these steps.
   1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/stereo-left-right}}, initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture using |context| , |stencil|, double of |width| and |height|.
   1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/stereo-top-bottom}}, initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture using |context| , |stencil|, |width| and double of |height|.
   1. return |array|.
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -928,7 +928,7 @@ To <dfn>allocate color textures for projection layers</dfn> using an {{XRProject
   1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/"mono"}} or {{XRLayerLayout/"default"}}:
         <dl class="switch">
         <dt> If |textureType| is {{"texture-array"}}:
-          <dd> Initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D_ARRAY}} texture with |numViews| internal textures using |context|, |alpha|, |width| and |height|.
+          <dd> Initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D_ARRAY}} texture with |numViews| layers using |context|, |alpha|, |width| and |height|.
           <dd> Return |array| and abort these steps.
         <dt> Otherwise
           <dd> For each |view| in the |session|'s [=list of views=]:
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -914,7 +915,66 @@ To <dfn>determine the layout attribute</dfn> using an {{XRTextureType}} |texture
...
-ISSUE: special case UA behavior if the size causes the layout to change (ie if the requested width exceeds a limit with {{XRLayerLayout/"stereo-left-right"}})
+<div class="algorithm" data-algorithm="allocate color textures for projection layers">
+
+To <dfn>allocate color textures for projection layers</dfn> using an {{XRProjectionLayer}} |layer|, an {{XRTextureType}} |textureType|, a boolean |alpha| and a float |scaleFactor|, the user agent MUST run the following steps:
+  1. let |array| be a [=new=] array in the [=relevant realm=] of |context|.
+  1. let |context| be |layer|'s [=XRCompositionLayer/context=].
+  1. let |session| be |layer|'s [=XRCompositionLayer/session=].
+  1. let |numViews| be the number of the |session|'s [=list of views=].
+  1. Let |view| be the first entry in the |session|'s [=list of views=].
+  1. Let |width| be the width of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
```
toji, line 926: Somewhere, maybe not here but at the site that calls this algorithm, we'll want to allow the UA to clamp the `scaleFactor`. Some devices won't be able to deviate from the recommended resolution, and most will have upper and lower bounds on the acceptable texture size. I don't see that as a blocking issue for this PR, though.

webxrlayers-1.bs
```diff
@@ -914,7 +915,66 @@ To <dfn>determine the layout attribute</dfn> using an {{XRTextureType}} |texture
...
+  1. Let |height| be the height of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
+  1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/"mono"}} or {{XRLayerLayout/"default"}}:
+        <dl class="switch">
+        <dt> If |textureType| is {{"texture-array"}}:
+          <dd> Initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D_ARRAY}} texture with |numViews| internal textures using |context|, |alpha|, |width| and |height|.
```
toji, line 931:
```suggestion
          <dd> Initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D_ARRAY}} texture with |numViews| layers using |context|, |alpha|, |width| and |height|.
```

"internal textures" -> "layers" to match WebGL's preferred terminology. It's a minor thing but it'll make it easier to cross reference between the two specs.

webxrlayers-1.bs
```diff
@@ -914,7 +915,66 @@ To <dfn>determine the layout attribute</dfn> using an {{XRTextureType}} |texture
...
+          <dd> Return |array| and abort these steps.
+        <dt> Otherwise
+          <dd> For each |view| in the |session|'s [=list of views=]:
+            1. Let |width| be the width of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
```
toji, line 935: Width and height are already computed outside the switch statement above with the same algorithm, so this is redundant. I think the width and height here should be different, though, to account for the fact that we're packing multiple views into a single layer. Given that we probably want to allow for UA/platform choice in how those are packed, though, I'm not sure exactly what this should look like.

webxrlayers-1.bs
```diff
@@ -914,7 +915,66 @@ To <dfn>determine the layout attribute</dfn> using an {{XRTextureType}} |texture
...
+  1. let |depthsupport| be true if |context| is a {{WebGL2RenderingContext}} or the {{WEBGL_depth_texture}} extension is enabled in |context|.
+  1. If |init|'s {{XRLayerInit/depth}} or {{XRLayerInit/stencil}} are set and |depthsupport| is false, throw {{TypeError}} and abort these steps.
+  1. let |numViews| be the number of the |session|'s [=list of views=].
+  1. Let |view| be the first entry in the |session|'s [=list of views=].
+  1. Let |width| be the width of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
+  1. Let |height| be the height of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
+  1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/"mono"}} or {{XRLayerLayout/"default"}}:
+        <dl class="switch">
+        <dt> If |textureType| is {{"texture-array"}}:
+          <dd> Initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D_ARRAY}} texture with |numViews| internal textures using |context|, |stencil|, |width| and |height|.
```
toji, line 963:
```suggestion
          <dd> Initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D_ARRAY}} texture with |numViews| layers using |context|, |stencil|, |width| and |height|.
```

webxrlayers-1.bs
```diff
@@ -914,7 +915,66 @@ To <dfn>determine the layout attribute</dfn> using an {{XRTextureType}} |texture
...
+          <dd> Return |array| and abort these steps.
+        <dt> Otherwise
+          <dd> For each |view| in the |session|'s [=list of views=]:
+            1. Let |width| be the width of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
```
toji, line 967: Same concern as width/height computation as above.

webxrlayers-1.bs
cabanier: I had looked at the WebXR spec but didn't see where it limited the `scaleFactor`. Can you point me where it is?

I will open an issue.

webxrlayers-1.bs
cabanier: The width and height can be different if we're allocating a texture per view since we want to allow views with different dimensions.
Maybe I should rename these variables?

webxrlayers-1.bs
toji: We should have one in the core spec but I'm not seeing it either right now. We do have a min/max clip plane that has very similar usage to what we'd want here: https://immersive-web.github.io/webxr/#minimum-near-clip-plane

We also have language in the explainer that indicates that the values may be clamped. https://github.com/immersive-web/webxr/blob/master/explainer.md#controlling-rendering-quality

webxrlayers-1.bs
```diff
@@ -914,7 +915,66 @@ To <dfn>determine the layout attribute</dfn> using an {{XRTextureType}} |texture
...
+            1. Let |height| be the height of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
+            1. let |texture| be a [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture with |context|, |alpha|, |width| and |height|.
+            1. Append |texture| to |array|.
+          <dd> Return |array| and abort these steps.
+        </dl>
+  1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/stereo-left-right}}, initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture using |context| , |alpha|, double of |width| and |height|.
+  1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/stereo-top-bottom}}, initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture using |context| , |alpha|, |width| and double of |height|.
```
asajeffrey, line 942: Reword both of these to something like "width `view[0].width + view[1].width` and height `max(view[0].height, view[1].height)`" (only said in standardsese). Either that or throw an exception if `views[0]` and `views[1]` are different sizes?

webxrlayers-1.bs
```diff
@@ -914,7 +915,66 @@ To <dfn>determine the layout attribute</dfn> using an {{XRTextureType}} |texture
...
+            1. Let |width| be the width of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
+            1. Let |height| be the height of |view|'s [=recommended WebGL texture resolution=] multiplied by |scaleFactor|.
+            1. let |texture| be a [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture with |context|, |stencil|, |width| and |height|.
+            1. Append |texture| to |array|.
+          <dd> Return |array| and abort these steps.
+        </dl>
+  1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/stereo-left-right}}, initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture using |context| , |stencil|, double of |width| and |height|.
+  1. If |layer|'s {{XRCompositionLayer/layout}} is {{XRLayerLayout/stereo-top-bottom}}, initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture using |context| , |stencil|, |width| and double of |height|.
```
asajeffrey, line 974: Ditto.

webxrlayers-1.bs
```diff
@@ -929,17 +989,14 @@ To <dfn>allocate color textures</dfn> using an {{XRCompositionLayer}} |layer|, a
...
+          <dd> Initialize |array| with 1 [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D_ARRAY}} texture with |numViews| layers using |context| and |init|'s {{XRLayerInit/alpha}}, {{XRLayerInit/viewPixelWidth}} and {{XRLayerInit/viewPixelHeight}} values.
           <dd> Return |array| and abort these steps.
         <dt> Otherwise
-          <dd> For each |view| in the |session|'s [=list of views=]:
-            1. Let |glViewport| be the [=WebGL viewport=] from the [=list of viewports=] associated with |view|.
-            1. let |texture| be a [=new=] instance of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture with |context|, |init|'s {{XRLayerInit/alpha}} and |glViewport|'s width and height.
-            1. Append |texture| to |array|.
+          <dd> Initialize |array| with |numViews| [=new=] instances of {{WebGLTexture}} in the [=relevant realm=] of |context| created as a {{TEXTURE_2D}} texture with |context| and |init|'s {{XRProjectionLayerInit/alpha}}, {{XRLayerInit/viewPixelWidth}} and {{XRLayerInit/viewPixelHeight}} values.
```
asajeffrey, line 999: This doesn't support different texture sizes for different views. Should it?

webxrlayers-1.bs
cabanier: o yes. We should throw when they are different sizes. I will fix.

webxrlayers-1.bs
cabanier: This is the algorithm for non-projection layers.
The size for the layer is given during the initialization and is the same for each view.

asajeffrey: Hmm, so for a device with a high-resolution HMD and a low-resolution camera, we're going to use the same texture size for all of them?

cabanier: yes

cabanier:
Also closes #162

cabanier:
Closes #162

AdaRoseCannon:
You may be interested, there is a bot command called "/fixes" so you can do "/fixes #162" and it adds a comment in that issue and adds a label that it's about to be fixed by a pending PR.

cabanier:
@toji @Manishearth this PR is getting a bit big. Can you review/approve it so I can update it for the observer view (if needed)

PR #165, opened 2020-06-16
Ensure non-projection layers can have at most 2 slots + added Eye parameter to getSubImageView

cabanier:
This change ensures that non-projection layers are no longer tied to views. 
It also changes getSubImageView so it only works on projection layers

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -250,6 +250,7 @@ The {{XRLayerLayout}} enum defines the layout of the layer.
 enum XRLayerLayout {
   "default",
   "mono",
+  "stereo",
   "stereo-left-right",
   "stereo-top-bottom"
 };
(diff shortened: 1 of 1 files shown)


cabanier:
Fixes #163

webxrlayers-1.bs
```diff
@@ -907,10 +910,12 @@ To <dfn>determine the layout attribute</dfn> using an {{XRTextureType}} |texture
...
   1. If |layout| is {{XRLayerLayout/"mono"}}, return |layout| and abort these steps.
   1. If |layout| is {{XRLayerLayout/"default"}}, run the following steps:
+    1. If the size of [=list of views=] is <code>1</code>, return {{XRLayerLayout/"mono"}} and abort these steps.
     1. If |textureType| is {{"texture-array"}}, return |layout| and abort these steps.
+  1. If |layout| is {{XRLayerLayout/"default"}} or {{XRLayerLayout/"stereo"}}, run the following steps:
     1. If the user agent prefers {{XRLayerLayout/"stereo-left-right"}} layout, return {{XRLayerLayout/"stereo-left-right"}} and abort these steps.
     1. If the user agent prefers {{XRLayerLayout/"stereo-top-bottom"}} layout, return {{XRLayerLayout/"stereo-top-bottom"}} and abort these steps.
-  1. If |textureType| is {{"texture-array"}}, throw {{TypeError}} and abort these steps.
+  1. If |textureType| is {{"texture-array"}} and |layout| is {{XRLayerLayout/"stereo-left-right"}} or {{XRLayerLayout/"stereo-top-bottom"}}, throw {{TypeError}} and abort these steps.
```
toji, line 918: I don't particularly oppose throwing in this situation, but wanted to point out that a possibly exception is returning a texture array with a single layer. Sounds silly, but based on the call today we may find ourselves doing that in some scenarios anyway? Don't block this PR on it, though.

webxrlayers-1.bs
```diff
@@ -859,8 +862,8 @@ interface XRWebGLBinding {
                                       optional XREquirectLayerInit init);
   XRCubeLayer createCubeLayer(optional XRCubeLayerInit init);
 
-  XRWebGLSubImage getSubImage(XRCompositionLayer layer, XRFrame frame);
-  XRWebGLSubImage getViewSubImage(XRCompositionLayer layer, XRView view);
+  XRWebGLSubImage getSubImage(XRCompositionLayer layer, XRFrame frame, optional XREye eye = "none");
+  XRWebGLSubImage getViewSubImage(XRProjectionLayer layer, XRView view);
```
toji, line 866: I really like the idea of explicitly limiting `getViewSubImage()` to `XRProjectionLayer`s! The only question it leaves me with is what happens if I pass a projection layer to `getSubImage()`? It's still an `XRCompositorLayer`, so the IDL won't reject it, but I don't think it's intended to work. We probably need a check in the algorithm below to reject it?

cabanier: I agree that it makes little sense that projection layers are allowed here.
I can't think of a good reason why an author would do that apart.

I will add the rejection of projection layers to the algorithm.

webxrlayers-1.bs
cabanier: I agree. We don't limit mono layers to texture_2d either. 
I will make the change.

PR #166, opened 2020-06-16
Add w3c.json to mark repo as host of rec-track work

dontcallmedom:


himorin:
looks good. (and I think this is the last fix for moving to WG)

Issue #193, opened 2020-08-06
XRProjectionLayer layers array behavior when the "layers" feature descriptor is not requested


MortimerGoro:
I'm implementing WebXR Layers in Firefox Reality and found this part of the spec confusing:
- Layers of type XRProjectionLayer MUST always be supported, regardless if the feature descriptor was requested.
- UpdateRenderState: if the layers array was passed and the session was not created with "layers" enabled, throw a NotSupportedError and abort these steps.

According to both statements the XRProjectionLayers can be created if the "layers" feature is not requested but can't be passed in the UpdateRenderState layers array if the layers feature is not enabled. So they are created for nothing.

We need to make this more clear in the spec. I guess the intended behavior is:  "if the layers array contains any XRLayer that is not a XRProjectionLayer or a XRWebGLLayer and the session was not created with "layers" enabled, throw a NotSupportedError and abort these steps"

asajeffrey:
Another fix would be to add a test to https://immersive-web.github.io/layers/#dom-xrwebglbinding-createprojectionlayer that throws a `NotSupportedException` if the session was created without `layers` enabled and `textureType` is not `texture`.

cabanier:
Yes. That was an oversight. I will fix this in the spec.
If `baselayer` is null, layers can have entries and vice versa.

cabanier:
@MortimerGoro I'm excited to hear that Firefox Reality is adding support!
Is that different from what @asajeffrey is working on?

Please let me know if you find any other problems in the spec.

MortimerGoro:
@cabanier yes, it's a different implementation. It's a long story, but Mozilla has two browser engines that are used in VR projects: Servo and Gecko/Firefox. I'm working on the Gecko implementation, which is the one used for Firefox Reality shipped on Oculus Quest/Go devices.

The spec looks good so far. I haven't read all the discussions, maybe it has been already commented: on a high level the only thing I miss would be a separation between a layer and the surface/texture (as OculusVR/OpenXR do with SwapChains). But that separation might not be a friendly web pattern. Example of a use case: In Firefox Reality frontend we used that separation to dynamically switch between a flat (Quad Layer) and curved (Cylinder Layer) window without recreating or painting to a different surface, so the switch is super fast. With WebXR Layers you would need to render to different textures.

cabanier:
[...]

PR #194, opened 2020-08-06
Fix up the pending layers state

cabanier:
Remove the reliance on 'layers' feature

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/194.html" title="Last updated on Aug 6, 2020, 7:11 PM UTC (bf0fc2d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/194/fa3a435...cabanier:bf0fc2d.html" title="Last updated on Aug 6, 2020, 7:11 PM UTC (bf0fc2d)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -1659,16 +1659,19 @@ updateRenderState changes {#updaterenderstatechanges}
 
 This module replaces the steps given by "[=update the pending layers state=]" from the WebXR specification. Instead when the user agent will
 <dfn dfn-for="layers">update the pending layers state</dfn> with {{XRSession}} |session| and {{XRRenderStateInit}} |newState|, it must run the following steps:
-  1. If |newState|'s {{XRRenderStateInit/baseLayer}} is set and the |session| was created with "[=feature descriptor/layers=]" enabled, throw a {{NotSupportedError}} and abort these steps.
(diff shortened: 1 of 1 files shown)


cabanier:
Closes #193

toji:
LGTM, but I wonder if a check for the "layers" feature ought to be added back in for array lengths > 1?

asajeffrey:
@toji if it makes any difference, Servo will soon support multiple projection and `XRWebGLLayer`s, but not quad, cylinder etc.

cabanier:
> LGTM, but I wonder if a check for the "layers" feature ought to be added back in for array lengths > 1?

Yes, right now we imply that any number of projection layers should be supported by default.
Should I open an issue?

toji:
I thought that in a previous discussion on one of the calls we decided that the "layers" feature would limit both non-projection layers _and_ multi-layer. We can re-evaluate that if we want, especially if it looks like the multi-layer part of it will be easy to support across the board.

cabanier:
> I thought that in a previous discussion on one of the calls we decided that the "layers" feature would limit both non-projection layers _and_ multi-layer. We can re-evaluate that if we want, especially if it looks like the multi-layer part of it will be easy to support across the board.

Yes, we discussed this before but I don't think this was decided.
Can you open an issue so we can discuss? This is likely not an issue for UA's that implement the full layers spec but it might be hard for others (like Magic Leap or cellphone AR)

PR #222, opened 2020-10-23
Allow creation of other texture types

### LINKED ISSUE #220: Return a list of supported color and depth texture types
We got some feedback from one of our users that they want to use a different texture format from RGB/RGBA.
Right now there is no way to query the list of supported layer formats or a way to request a texture type.

Should we give this ability in the layers API? I see that @toji proposed something similar for [WebGPU Layers](https://github.com/toji/webxr-webgpu-binding/blob/main/explainer.md).

cabanier:
Closes #220

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/222.html" title="Last updated on Oct 27, 2020, 1:59 AM UTC (b7e9045)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/222/d6e6475...cabanier:b7e9045.html" title="Last updated on Oct 27, 2020, 1:59 AM UTC (b7e9045)">Diff</a>


webxrlayers-1.bs
```diff
@@ -769,21 +771,26 @@ is initialized.
 <pre class="idl">
 dictionary XRProjectionLayerInit {
   XRTextureType textureType = "texture";
+  GLenum colorFormat = 0x1907; // RGBA
+  GLenum depthFormat = 0x84FA; // UNSIGNED_INT_24_8_WEBGL
```
toji, line 775: Here and below, shouldn't this more accurately be `depthStencilFormat`?

Also, it would be nice if we could use this to eliminate the `depth` and `stencil` booleans in the same way that specifying the `colorFormat` implicitly eliminates the need for `alpha`. It's a bit awkward because it's a `GLenum`, but we could either make it nullable or specify that passing `gl.NONE` (0) indicates that no depth/stencil buffer needs to be provided. What do you think?

webxrlayers-1.bs
```diff
@@ -769,21 +771,26 @@ is initialized.
...
 };
 </pre>
 
 The <dfn dict-member for="XRProjectionLayerInit">textureType</dfn> attribute defines the type of texture that the layer will have.
 
+The <dfn dict-member for="XRProjectionLayerInit">colorFormat</dfn> attribute defines the data type of the color texture data.
+The default value is {{RGBA}}.
+
+The <dfn dict-member for="XRProjectionLayerInit">depthFormat</dfn> attribute defines the data type of the depth texture data.
+The default value is {{UNSIGNED_INT_24_8_WEBGL}}.
```
toji, line 788: Hm... There's a couple of issues with the default enum here.

For one, this isn't a base WebGL 1 enum, but that makes me realize that WebGL 1 doesn't actually support depth textures without an extension. Yuck. 😫 I'm going to file a separate issue on that, because we'll have to address it no matter what approach we take here.

Second, ignoring the first issue, this changes the defaults a bit in that it implies stencil is present, which the original booleans default to false. Granted, until you hit WebGL2 there's not a way to require 24 bit depth without stencil, and the precision loss doesn't make for a good default.

Kinda wondering if the right approach is to actually default to `gl.NONE`, let WebGL 1 baseline users use their own renderbuffers, and force devs to make an explicit choice about which format they use otherwise. It'll be a bit of a departure from the patterns we currently have set up, but if anything now would be the time to make those kind of breaking changes. (I'd be happy to reflect any default patterns we establish here in the WebGPU layers for consistency.)

webxrlayers-1.bs
```diff
@@ -898,6 +910,9 @@ The {{XRWebGLBinding}} object is used to create layers that have a GPU backend.
 
   XRWebGLSubImage getSubImage(XRCompositionLayer layer, XRFrame frame, optional XREye eye = "none");
   XRWebGLSubImage getViewSubImage(XRProjectionLayer layer, XRView view);
+
+  readonly attribute FrozenArray&lt;GLenum&gt; supportedColorFormats;
+  readonly attribute FrozenArray&lt;GLenum&gt; supportedDepthStencilFormats;
```
toji, line 915: 👍 Having these as readonly attributes feels better than the getter functions I proposed for the WebGPU bindings.  I'll make the same change there.

webxrlayers-1.bs
```diff
@@ -769,21 +771,26 @@ is initialized.
 <pre class="idl">
 dictionary XRProjectionLayerInit {
   XRTextureType textureType = "texture";
+  GLenum colorFormat = 0x1907; // RGBA
```
toji, line 774: Any concerns that the default here may not align with the advertised preferred format? I was considering doing the same thing with WebGPU layers, but was a bit concerned about both requiring a specific format and choosing an arbitrary one that may not be optimal. That said, it does make layer init a lot easier for simply cases if there is a default here.

toji (commented):
I like this approach overall, since it allows more developer control and better aligns with what the WebGPU layers will be required to do. A few points I'd like to discuss further below:

webxrlayers-1.bs
cabanier: Are there platforms where plain RGBA is not the preferred format?
Do you think we should have an attribute that returns the preferred format (and then default to that in the absence of the colorFormat entry in the init structure)?

webxrlayers-1.bs
cabanier: > Here and below, shouldn't this more accurately be `depthStencilFormat`?

I agree. I will make that change.

> Also, it would be nice if we could use this to eliminate the `depth` and `stencil` booleans in the same way that specifying the `colorFormat` implicitly eliminates the need for `alpha`. It's a bit awkward because it's a `GLenum`, but we could either make it nullable or specify that passing `gl.NONE` (0) indicates that no depth/stencil buffer needs to be provided. What do you think?

I *think* it makes more sense to make it nullable .

webxrlayers-1.bs
cabanier: > For one, this isn't a base WebGL 1 enum, but that makes me realize that WebGL 1 doesn't actually support depth textures without an extension. Yuck. 😫 I'm going to file a separate issue on that, because we'll have to address it no matter what approach we take here.

The spec already deals with that. For instance see step 4  in [this](https://immersive-web.github.io/layers/#allocate-depth-textures) algorithm.

> Second, ignoring the first issue, this changes the defaults a bit in that it implies stencil is present, which the original booleans default to false. Granted, until you hit WebGL2 there's not a way to require 24 bit depth without stencil, and the precision loss doesn't make for a good default.
> 
> Kinda wondering if the right approach is to actually default to `gl.NONE`, let WebGL 1 baseline users use their own renderbuffers, and force devs to make an explicit choice about which format they use otherwise. It'll be a bit of a departure from the patterns we currently have set up, but if anything now would be the time to make those kind of breaking changes. (I'd be happy to reflect any default patterns we establish here in the WebGPU layers for consistency.)

Maybe we can get rid of the depth and stencil booleans in the init structure and then rely on the enum to determine if depth or stencil are requested.

webxrlayers-1.bs
toji: It certainly feels more "webby". Easier to infer the intent from reading the IDL that way too.

webxrlayers-1.bs
toji: > The spec already deals with that. For instance see step 4 in this algorithm.

Awesome! Didn't see that earlier (and it mostly makes #223 moot.) I do think we should re-evaluate the defaults, since the current setup would cause it to always fail on baseline WebGL 1 contexts.

> Maybe we can get rid of the depth and stencil booleans in the init structure and then rely on the enum to determine if depth or stencil are requested.

Yeah, I think that feels best.

webxrlayers-1.bs
toji: > Are there platforms where plain RGBA is not the preferred format?

I think BGRA is preferred on some devices? Regardless, I doubt it makes a huge difference performance difference for rendering? It's probably a bigger concern when doing texture copies.

> Do you think we should have an attribute that returns the preferred format

My intent with WebGPU layers was to follow OpenXR's convention of having the supported list be returned in order of preference.

> (and then default to that in the absence of the colorFormat entry in the init structure)

The idea of defaulting to the preferred format if the `colorFormat` is `null` is intriguing, but I worry that it would create confusion when paired with the `depthStencilFormat` (where we're discussing having `null` mean "Don't allocate a texture"). Also, in some cases where the developer is directly providing image data via `texSubImage2d()` or similar having the format shift out from under them may lead to situations where their code works on one device but shows "corrupted" image data on another that they didn't test on. In that regard a stable default, even if not the optimal format, is probably better.

So yeah, I think I just talked myself out of my concerns. And RGBA certainly seems like a logical default.

toji (approved):
Looks great! Thanks!

PR #229, opened 2020-11-18
Removed supportedColorFormats and supportedDepthStencilFormats. Replaced with hardcoded list.

### LINKED ISSUE #228: Should we limit the list of supported texture formats?
Implementing the list of supported formats is showing a large amount of texture color and depth formats.
Is it really necessary to surface them all and allow them in the API?

My worry is that this might create content that only works on certain devices (ie Quest supports a format and Hololens doesn't) and that this might leak information.

/agenda discuss if we should limit the list of supported texture formats

cabanier:
Closes #228

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/229.html" title="Last updated on Nov 21, 2020, 12:13 AM UTC (ed09308)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/229/36946c4...cabanier:ed09308.html" title="Last updated on Nov 21, 2020, 12:13 AM UTC (ed09308)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -896,7 +896,7 @@ For WebGL1 contexts with the '{{WEBGL_depth_texture}}' [=extension=] enabled or 
  - {{DEPTH_COMPONENT}}
  - {{DEPTH_STENCIL}}
 
-For WebGL2 contexts these additional formats are supported::
+For WebGL2 contexts these additional formats are supported:
 
  - {{DEPTH_COMPONENT24}}
  - {{DEPTH24_STENCIL8}}
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -90,11 +90,38 @@ spec: WebGL; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.0/
     type: dfn; text: WebGL viewport; url:#5.14.4
     type: typedef; text: GLenum; url: #5.1
     type: typedef; text: RGBA; url: #5.14
+    type: typedef; text: RGB; url: #5.14
+    type: dfn; text: extension; url:#5.14.14
+spec: ; urlPrefix: https://www.khronos.org/registry/webgl/extensions/EXT_sRGB/
+    type: typedef; text: SRGB_EXT
+    type: typedef; text: SRGB_ALPHA_EXT
+    type: typedef; text: EXT_sRGB
+spec: ; urlPrefix: https://www.khronos.org/registry/webgl/extensions/WEBGL_depth_texture/
+    type: typedef; text: DEPTH_COMPONENT
+    type: typedef; text: DEPTH_STENCIL
+    type: typedef; text: WEBGL_depth_texture
+spec: ; urlPrefix: https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_etc/
+    type: typedef; text: COMPRESSED_RGB8_ETC2
+    type: typedef; text: COMPRESSED_RGB8_PUNCHTHROUGH_ALPHA1_ETC2
+    type: typedef; text: COMPRESSED_RGBA8_ETC2_EAC
+    type: typedef; text: COMPRESSED_SRGB8_ETC2
+    type: typedef; text: COMPRESSED_SRGB8_PUNCHTHROUGH_ALPHA1_ETC2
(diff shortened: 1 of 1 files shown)


cabanier:
@toji how does this look?

webxrlayers-1.bs
```diff
@@ -811,11 +863,44 @@ dictionary XRLayerInit {
...
+
+For contexts with the '{{WEBGL_compressed_texture_etc}}' [=extension=] enabled these additional formats are supported:
+ - {{COMPRESSED_RGB8_ETC2}}
+ - {{COMPRESSED_RGB8_PUNCHTHROUGH_ALPHA1_ETC2}}
+ - {{COMPRESSED_RGBA8_ETC2_EAC}}
+ - {{COMPRESSED_SRGB8_ETC2}}
+ - {{COMPRESSED_SRGB8_PUNCHTHROUGH_ALPHA1_ETC2}}
+ - {{COMPRESSED_SRGB8_ALPHA8_ETC2_EAC}}
+
+For contexts with the '{{WEBGL_compressed_texture_astc}}' [=extension=] enabled all the formats of that [=extension=] are supported.
```
toji, line 887: Is there a reason why only these two compressed texture extensions are called out? I would be surprised, for example, if S3TC wasn't supported on desktop systems.

webxrlayers-1.bs
```diff
@@ -811,11 +863,44 @@ dictionary XRLayerInit {
...
 If {{XRLayerInit/depthFormat}} is not supplied, the layer will not provide a depth/stencil texture.
 
+This is the <dfn>list of depth formats for non-projection layers</dfn> that the [=XR Compositor=] MUST support:
+
+For WebGL1 contexts with the '{{WEBGL_depth_texture}}' [=extension=] enabled or WebGL2 contexts:
+
+ - {{DEPTH_COMPONENT}}
+ - {{DEPTH_STENCIL}}
+
+For WebGL2 contexts these additional formats are supported::
```
toji, line 899:
```suggestion
For WebGL2 contexts these additional formats are supported:
```

toji (commented):
Looks great overall, but two questions I'd like answered prior to merging.

webxrlayers-1.bs
```diff
@@ -775,18 +802,43 @@ is initialized.
 dictionary XRProjectionLayerInit {
   XRTextureType textureType = "texture";
   GLenum colorFormat = 0x1908; // RGBA
-  GLenum? depthFormat;
+  GLenum depthFormat = 0x1902; // DEPTH_COMPONENT
```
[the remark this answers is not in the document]
cabanier: What do you mean by "the default"?
You are right that for WebGL1, this needs the extension. This was already the case in the previous approach with the booleans because projection layers were always created with a depth buffer.

webxrlayers-1.bs
cabanier: We wanted to create lists of formats that were supported by all systems.  S3TC is not supported on Qualcomm.

Maybe we should revisit and state: "if the extension can be enabled, then the following formats are supported..."

webxrlayers-1.bs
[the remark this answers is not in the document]
toji: In the previous approach the depth attachment could be supplied by using a Renderbuffer, which we can't do here. I'd feel more comfortable is the default was no depth buffer.

webxrlayers-1.bs
toji: That sounds like a good approach, though I'd say that the current verbiage works fine for the scenario where the extension isn't available as well, since you obviously can't enable it in that case.

cabanier: @thetuvix Do you have an opinion? Should we add formats that everyone supports or should we let it depend on the extension + the formats of that extension that make sense (ie no single component or depth for the swapchain)

webxrlayers-1.bs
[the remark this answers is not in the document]
cabanier: You can still do that. Nothing has changed in that regard.
If you pass a value of `0`, no depth buffer is created and you can create your own. I agree that feels a bit weird but I think a projection layer should have a depth buffer by default.

toji (approved):
I'm fine with merging this at this point.

PR #241, opened 2021-01-15
Texture size attributes

### LINKED ISSUE #224: Communicate projection layer texture size outside the frame loop
Brought up in discussion on #223 

Most layer types accept a user-specified texture size, but projection layers only accept a scalar and the final size is determined by the UA/hardware. Currently this size is only reported in the `XRSubImage`, which can only be accessed inside the frame loop.

The problem is that currently if the application wants to manage their own depth buffer (which will be necessary for WebGL 1 apps that aren't using the `WEBGL_depth_texture` extension or that don't want their depth buffer to be used by the compositor), or any other supplemental buffers, it would have to wait until the first frame after layer creation to determine the required dimensions for those buffers. Ideally that allocation could happen outside of the frame loop, probably immediately after the layer creation.

I can think of one of two ways to approach this:

1) We report the texture dimensions on the `XRProjectionLayer` directly. (This would likely be ONLY for the projection layer, since other layer types may be used with things like `XRMediaBinding` and the sizes would be moot.) In order for applications to allocate a matching buffer we'd want to report both the size and layer count for array textures.

```webidl
[Exposed=Window] interface XRProjectionLayer : XRCompositionLayer {
  readonly attribute unsigned long textureWidth;
  readonly attribute unsigned long textureHeight;
  readonly attribute unsigned long textureLayers;
};
```

2) If we want this to work a bit more generally and sidestep complications with `XRMediaBinding` we could also make it a query on `XRWebGLBinding`, similar to `getSubImage()` but without the frame requirements:

```webidl
partial interface XRWebGLBinding {
   XRImageDimensions getImageDimensions(XRCompositionLayer layer);
};

interface XRImageDimensions {
  readonly attribute unsigned long textureWidth;
  readonly attribute unsigned long textureHeight;
  readonly attribute unsigned long textureLayers;
};
// Probably want to use that in XRWebGLSubImage too?
```

cabanier:
Closes #224

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/241.html" title="Last updated on Jan 15, 2021, 7:42 PM UTC (769c0fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/241/bab6191...cabanier:769c0fe.html" title="Last updated on Jan 15, 2021, 7:42 PM UTC (769c0fe)">Diff</a>


toji (approved):
LGTM!

PR #243, opened 2021-01-21
Add support for mipmapping

### LINKED ISSUE #237: Support mipmapping in composition layers
We currently don't support mipmapping.
It seems that it would be useful for experiences that want to display high quality content.

I don't think it's useful for projection layers because they are always shown at full pixel resolution.
Maybe it also doesn't make sense for cube and equirect layers.

cabanier:
Closes #237

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/243.html" title="Last updated on Jan 22, 2021, 8:59 PM UTC (ff5e235)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/243/f2ad12b...cabanier:ff5e235.html" title="Last updated on Jan 22, 2021, 8:59 PM UTC (ff5e235)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -822,6 +822,9 @@ The buffers attached to an [=opaque texture=] MUST be cleared to the values in t
   </tbody>
 </table>
 
+If the [=opaque texture=] was created with <code>2</code> or more {{XRLayerInit/mipLevels}},
+the author SHOULD populate all the mip levels. The user agent MUST NOT assume that it should create the mip levels.
+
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM! Only question is, do we want to leave a note in here (even if non-normative) that indicates that user agents can perform automatic mip-mapping on projection layers if desired?

cabanier:
> LGTM! Only question is, do we want to leave a note in here (even if non-normative) that indicates that user agents can perform automatic mip-mapping on projection layers if desired?

I was on the fence about adding this. Since you brought it up, I will add a non-normative note.

PR #245, opened 2021-01-28
Add support to return the number of created mip levels

cabanier:
***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/245.html" title="Last updated on Jan 28, 2021, 9:15 AM UTC (d3a3c75)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/245/5639438...cabanier:d3a3c75.html" title="Last updated on Jan 28, 2021, 9:15 AM UTC (d3a3c75)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -319,6 +319,7 @@ XRCompositionLayer defines a set of common attributes and behaviors across certa
   attribute boolean blendTextureSourceAlpha;
   attribute boolean? chromaticAberrationCorrection;
   attribute float? fixedFoveation;
+  readonly attribute unsigned long mipLevels;
 
   readonly attribute boolean needsRedraw;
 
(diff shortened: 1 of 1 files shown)


toji (approved):
This looks good, thanks!

I was weighing the pros/cons of this approach vs advertising a maxMipLevels and erroring if the developer requests more, and I don't think there's a significant difference either way. That approach would surface errors earlier, and they'd be WebXR errors rather than WebGL errors, but I don't think that actually makes much difference? Neither approach prevents developers testing on a system that support a full mip chain from accidentally writing code that errors on systems that don't.

The most common scenarios I see happening are: developer calls `gl.generateMipmap()` which does the right thing automatically no matter mow many mip levels there are, or they do a for loop through each level to upload something. In that case if we provide the mipLevels on the layer itself they're more likely to use that, which would offer some slight protection from error cases. (And in WebGL if you try to upload to mip levels that don't exist it'll just error but not fundamentally break anything. WebGPU will be trickier in this regard.)

So all of that is a long way of saying: This approach seems like the better option! LGTM!

cabanier:
Thanks @toji !

PR #250, opened 2021-02-19
Allow creation of projection layers if depth support isn't enabled in the context

### LINKED ISSUE #249: Not possible to create a default projectionlayer on a default WebGL Context
The spec currently checks if depth support is enabled in the context before allowing the creation of depth textures.
As a result, it's currently not possible to create a default projection layer because it has depth support enabled by default.

I proposed that we ignore the request for depth support if it's not enabled in the context, but only for projection layers.

cabanier:
Closes #249

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/250.html" title="Last updated on Feb 19, 2021, 7:34 PM UTC (8337ae8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/250/a8346f5...cabanier:8337ae8.html" title="Last updated on Feb 19, 2021, 7:34 PM UTC (8337ae8)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -874,12 +874,16 @@ If {{XRProjectionLayerInit/depthFormat}} is <code>0</code> the layer will not pr
 
 This is the <dfn>list of depth formats for projection layers</dfn> that the [=XR Compositor=] MUST support:
 
-For WebGL1 contexts with the '{{WEBGL_depth_texture}}' [=extension=] enabled or WebGL2 contexts:
+For {{WebGLRenderingContext}} contexts with the '{{WEBGL_depth_texture}}' [=extension=] enabled or WebGL2 contexts:
 
  - {{DEPTH_COMPONENT}}
  - {{DEPTH_STENCIL}}
 
-For WebGL2 contexts these additional formats are supported:
+If the extension was not enabled, the request for a depth texture is ignored.
+
+NOTE: this could be confusing to authors because they might expect a depth texture. If possible, provide a warning with the reason why the texture was not created.
+
+For {{WebGL2RenderingContext}} contexts these additional formats are supported:
 
  - {{DEPTH_COMPONENT24}}
  - {{DEPTH24_STENCIL8}}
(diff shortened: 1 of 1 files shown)


toji:
So to be clear: this change makes it so that if you request a depth texture (which is still the default) but don't have the depth extension enabled for a WebGL 1 context then you'll get back a layer with only the color texture.

That works, though it feels a bit weird. Tried thinking through the worst case scenarios:

Worst case is that you didn't enable the extension and question why you didn't get a depth buffer when the default says you should have. In that case the browser should definitely be outputting a warning to explain what happened. We can't put that in the spec text, but maybe a non-normative note suggesting it would be good. And in any case, that's not an error that you're likely to ship with.

More problematic is that the object returned by enabling the extension only has an enum on it, so it's easy to ignore the result and simply assume that the extension was enabled. So you may request the extension, assume it works, and observe that the code runs on your machine only to find that it fails on a different device. HOWEVER! It's **extremely** unlikely that any device that can even pretend to support VR won't have the ability to do depth textures, so this is more of a theoretical issue than a real world concern.

Finally, there's a the issue that someone may not enable the extension initially and depend on this behavior to prevent allocation of the depth buffer (which they're not using) only to later enable the extension and suddenly be allocating an unnecessary texture. This would be further compounded by the fact that we assume that if a depth texture is allocated we can use it for compositing. I think this is probably a rare situation, but it's worth bringing up.

That said, I don't think the above concerns are worth blocking this change for. I simply wanted to list them out to acknowledge that we thought about them and/or give you an opportunity to address them if you happen to have an idea about how to address it.

toji (approved):

cabanier:
> Worst case is that you didn't enable the extension and question why you didn't get a depth buffer when the default says you should have. In that case the browser should definitely be outputting a warning to explain what happened. We can't put that in the spec text, but maybe a non-normative note suggesting it would be good. And in any case, that's not an error that you're likely to ship with.

Good point!
I'll add a note to the spec and when I implement it, I will make sure to log a warning to the console.

> That said, I don't think the above concerns are worth blocking this change for. I simply wanted to list them out to acknowledge that we thought about them and/or give you an opportunity to address them if you happen to have an idea about how to address it.

Thanks!

toji:
Perfect, thanks!

PR #269, opened 2021-10-22
Added support for ignoreDepthValues to XRWebGLBinding

cabanier:
Closes https://github.com/immersive-web/webxr/issues/1228

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/269.html" title="Last updated on Nov 1, 2021, 6:16 PM UTC (74e2662)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/269/3aa5fc1...cabanier:74e2662.html" title="Last updated on Nov 1, 2021, 6:16 PM UTC (74e2662)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -1040,6 +1040,7 @@ The {{XRWebGLBinding}} object is used to create layers that have a GPU backend.
   constructor(XRSession session, XRWebGLRenderingContext context);
 
   readonly attribute double nativeProjectionScaleFactor;
+  readonly attribute boolean usesDepthValues;
 
   XRProjectionLayer createProjectionLayer(optional XRProjectionLayerInit init = {});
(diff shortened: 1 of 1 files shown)


cabanier:
> That said, during the TPAC call we also discussed that it may be more useful to communicate that a system prefers to have depth values provided, so perhaps we should consider flipping the logic and using a name like `prefersDepthValues` or, more generally, `usesDepthValues`.

Thanks! I updated the spec so it reads `usesDepthValues`

cabanier:
> I could go either way on that. It's nice to have them be aligned but the context is a bit different and the attrib in the XRProjectionLayer has already shipped in at least one browser, right? I don't think that makes it unchangeable, but I also don't feel the need to break things in this case.

Yes, we already shipped it and three is checking it.

toji (approved):

PR #271, opened 2021-11-01
Add support for depth testing across layers

### LINKED ISSUE #135: Depth testing across layers
The explainer mentions depth testing is not to be performed across layers. What is the reasoning behind that? Is it because the platforms targeted currently do not support depth buffers for projection layers?

Also, could it be allowed as an optional feature now or in the future?

cabanier:
Closes #135

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/271.html" title="Last updated on Nov 2, 2021, 11:02 PM UTC (66c8a90)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/271/3aa5fc1...cabanier:66c8a90.html" title="Last updated on Nov 2, 2021, 11:02 PM UTC (66c8a90)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -1875,6 +1875,18 @@ when the [=XR Compositor=] can no longer reproject the layer.
 
 The author SHOULD redraw the content of the layer at the next [=XR animation frame=]. The event must be of type {{XRLayerEvent}}.
 
+Depth sorting between layers {#depthsorting}
+============================
+By default, the {{XRRenderStateInit/layers}} array defines the order of the composition of the layers and each layer is drawn on top of the previous layer.
+If an application wants to have layers that are sorted by depth, it MUST request a session with the "<dfn for="feature descriptor">depth-sort</dfn>" [=feature descriptor=].
+
+If an {{XRSession}} is created with the "[=depth-sort=]" [=feature descriptor=], {{XRProjectionLayer}}, {{XRQuadLayer}} and {{XRCylinderLayer}} layers MUST be displayed based on their depth as opposed to the location in the {{XRRenderStateInit/layers}} array.
+Other layers types MUST continue to be sorted as before.
+
+{{XRQuadLayer}} and {{XRCylinderLayer}} layers MUST be sorted by their dimensions (for instance {{XRQuadLayerInit/width}} or {{XRCylinderLayerInit/centralAngle}}), transform and space.
+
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -1875,6 +1875,18 @@ when the [=XR Compositor=] can no longer reproject the layer.
 
 The author SHOULD redraw the content of the layer at the next [=XR animation frame=]. The event must be of type {{XRLayerEvent}}.
 
+Depth sorting between layers {#depthsorting}
+============================
+By default, the {{XRRenderStateInit/layers}} array defines the order of the composition of the layers and each layer is drawn on top of the previous layer.
+If an application wants to have layers that are sorted by depth, it MUST request a session with the "<dfn for="feature descriptor">depth-sort</dfn>" [=feature descriptor=].
+
+If an {{XRSession}} is created with the "[=depth-sort=]" [=feature descriptor=], {{XRProjectionLayer}}, {{XRQuadLayer}} and {{XRCylinderLayer}} layers MUST be displayed based on their depth as opposed to the location in the {{XRRenderStateInit/layers}} array.
```
toji, line 1883: I would prefer that the feature descriptor contain the term "layer" somewhere, because otherwise "depth-sort" feels a bit too generic.

"depth-sort-layers" or "layer-depth-sorting" perhaps?

toji (commented):
I'd like to make a general request to do some research and see if there's a specific use case that drove the OpenXR design of the same feature, which allows depth sorting to be toggled per-layer, but otherwise this seems OK with one naming nit.

cabanier:
> I'd like to make a general request to do some research and see if there's a specific use case that drove the OpenXR design of the same feature, which allows depth sorting to be toggled per-layer, but otherwise this seems OK with one naming nit.

I asked internally and that flag is so the system can make sure that quad/cylinder layers can always be drawn on top of the scene. 

For instance, if you hit the Oculus button during an immersive session, the dialog that then pops up should always take precedence over what the browser is drawing. Same with the keyboard which should always be in front.
I don't think that's a particular concern for WebXR.

toji:
Agreed. Thanks for looking into it!

toji (approved):

eXponenta:
Looks like that in Oculus2 there are a flag, but still not working.
I'm tried to use it - session starts, but quad layers not sorted between themselfs, need to sort manually by Z across camera.

cabanier:
Yes, there is a flag but the feature is not enabled yet. The flag will just trigger all the logic in the browser but the OS won't honor it.
We'll announce when it's ready for experimentation.

PR #285, opened 2022-07-07
Tie XRLayerEvent to redraw event when fired

dontcallmedom:
Align with DOM suggesting phrasing


cabanier:
Thanks!

cabanier:
@dontcallmedom there seems to be a build problem

himorin:
@cabanier might be due to overlapping of defined charter period.. let me talk internally

PR #305, opened 2024-02-08
Clarify rgb/srgb behavior

cabanier:
cc @klausw
See draft spec at  https://cabanier.github.io/layers/#xropaquetextures

addresses [#988](https://github.com/immersive-web/webxr/issues/988)

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/305.html" title="Last updated on Feb 9, 2024, 3:41 AM UTC (805f743)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/305/b79f6a1...cabanier:805f743.html" title="Last updated on Feb 9, 2024, 3:41 AM UTC (805f743)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -54,6 +54,7 @@ spec: webxr;
     type: dfn; text: primary view
     type: dfn; text: secondary view
     type: dfn; text: active; for: view
+    type: enum-value; text: local
 spec: html;
     type: dfn; text: check the usability of the image argument
     type: dfn; text: request the xr permission
@@ -460,7 +461,7 @@ XRProjectionLayer {#xrprojectionlayertype}
 An {{XRProjectionLayer}} is a layer that fills the entire view of the observer.
 Projection layers should be refreshed close to the device's native frame rate.
 
-<img alt="representation of a projection layer" src="images/projection-layer.jpg" style="width:50%; height: 50%; border-style: ridge;"/>
+<img alt="representation of a projection layer" src="images/projection-layer.jpg" style="width:50%; height: 50%; border-style: ridge;">
 
 <pre class="idl">
 [Exposed=Window] interface XRProjectionLayer : XRCompositionLayer {
@@ -518,7 +519,7 @@ Only the front of the layer MUST be visible; the back face MUST not be drawn by 
 A XRQuadLayer has no thicknes. It is a two-dimensional object positioned and oriented in 3D space. The position
 of a quad refers to the center of the quad.
 
-<img alt="representation of a quad layer" src="images/quad-layer.jpg" style="width:50%; height: 50%; border-style: ridge;"/>
+<img alt="representation of a quad layer" src="images/quad-layer.jpg" style="width:50%; height: 50%; border-style: ridge;">
 
 <pre class="idl">
 [Exposed=Window] interface XRQuadLayer : XRCompositionLayer {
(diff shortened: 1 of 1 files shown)


toji:
This matches my understanding of what the base WebGL contexts are doing, and I'm strongly in favor of the color output being consistent between the 2D browser and immersive modes.

webxrlayers-1.bs
```diff
@@ -908,6 +909,11 @@ section in the OpenGL ES 3.0 spec defines these limitations in more detail.
 
 Allocation of the resources for layers (such as memory) MUST be done through the same mechanism as WebGL.
 
+If an {{XRLayer}} is allocated with the {{RGBA}} or {{RGB}} {{XRLayerInit/colorFormat}}, its [=colorTextures=] MUST be exposed as {{RGBA}} or {{RGB}} to the {{WebGLRenderingContext}} context.
```
klausw, line 912: Maybe add a note here that RGBA is the default, and this is equivalent to the behavior of using a XRWebGLLayer without the Layers extension (using its default alpha=true).

cabanier: The spec already calls out that RGBA is the default where it's defining the initialization structures.

webxrlayers-1.bs
```diff
@@ -908,6 +909,11 @@ section in the OpenGL ES 3.0 spec defines these limitations in more detail.
...
+However, the [=XR Compositor=] MUST treat the [=colorTextures=]'s pixels as if they were in the {{SRGB8_ALPHA8}} or {{SRGB8}} {{XRLayerInit/colorFormat}}.
+
+NOTE: this means that the [=XR Compositor=] MUST not do any gamma conversion from linear {{RGBA}} or {{RGB}} when it processes the [=colorTextures=]. Otherwise, the pixels in the final rendering will appear too bright which will not match the rendering on a regular 2D {{WebGLRenderingContext}} context.
```
klausw, line 915: s/MUST not/MUST NOT/

There's a second problem also if the compositor erroneously treats the data as linear for compositing, but I'm not sure how to best fit that into the note. Maybe something like this?

NOTE: this means that the [=XR Compositor=] MUST NOT do any gamma conversion of the [=colorTextures=] pixel values *from* linear {{RGBA}} or {{RGB}} when it processes them for output on a device expecting nonlinear sRGB data. Conversely, if the [=XR Compositor=] requires linear data for blending or other internal processing, it MUST do a reverse gamma conversion of the [=colorTextures=] pixel values *to* linear. Otherwise, the pixels in the final rendering will appear too bright which will not match the rendering on a regular 2D {{WebGLRenderingContext}} context.

klausw (commented):
Looks good overall, thank you!

webxrlayers-1.bs
cabanier: I updated the upper casing of "MUST NOT" everywhere in the document.
I'm not aware of any compositor that works in linear light so I'm unsure if we should add it. We can discuss in the next meeting (or the face to face if you want to join).

klausw: Technically, even simple interpolation or averaging from a GL_LINEAR texture lookup, for example when reading the projection layer during reprojection, will get the wrong result unless the calculation happens in linear space. As far as I understand, the GPU pretty much needs to know the correct color encoding for anything other than a 1:1 pixel copy. (Apparently the needed conversions are essentially free on modern GPUs?)

But this is a minor thing, it's implied from the new spec text. I can try to call in to the discussion if that works out, but I think we now have something solid already and I will try to refrain from obsessing over the remaining details....

PR #312, opened 2025-03-25
Add support for foveateBoundTexture +  small bikeshed nits

cabanier:
First pass at adding support for foveateBoundTexture. 
I still need to investigate how I can amend the spec to remove the foveation.


webxrlayers-1.bs
```diff
@@ -1764,6 +1770,20 @@ When this method is invoked on an {{XRWebGLBinding}} |binding|, it MUST run the
...
+The <dfn method for="XRWebGLBinding">foveateBoundTexture(GLenum |target|, float |fixed_foveation|)</dfn> applies foveation to the current bound texture or texture array. 
+
+When this method is invoked on an {{XRWebGLBinding}} |binding|, it MUST run the following steps:
+
+  1. If |target| is not of type {{TEXTURE_2D}} or {{TEXTURE_2D_ARRAY}}, throw an {{InvalidStateError}} and abort these steps.
+  1. If the user agent or device does not support foveation, abort these steps.
+  1. If |fixed_foveation| is less than <code>0</code>, set it to <code>0</code>.
+  1. If |fixed_foveation| is more than <code>1</code>, set it to <code>1</code>.
+  1. If no texture is bound to |target| in |binding|'s [=XRWebGLBinding/context=], throw an {{InvalidStateError}} and abort these steps.
+  1. Apply |fixed_foveation| to |target| using |binding|'s [=XRWebGLBinding/context=].
```
toji, line 1782: I feel like this step needs more detail, but you probably already knew that. :)

Specifically, we should be specifying the relationship between the float and how much foveation is applied (is 0 the highest quality or 1? Does setting the highest quality imply that foveation is removed?), make it clear what aspects of foveation are determined by the implementation, etc.

cabanier: Maybe I can refer to the WebXR spec's definition of fixedFoveation.

toji: That would probably be sufficient, yeah.

toji (approved):
LGTM, thanks for the updates! (And yeah, the target arg is necessary, thanks for catching that.)

PR #334, opened 2026-05-31
clarify spacewarp and mark as no longer unstable

cabanier:
***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/layers/pull/334.html" title="Last updated on Jun 1, 2026, 6:28 PM UTC (d8988fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/layers/334/0d61133...cabanier:d8988fa.html" title="Last updated on Jun 1, 2026, 6:28 PM UTC (d8988fa)">Diff</a>

### DIFF
--- a/webxrlayers-1.bs
+++ b/webxrlayers-1.bs
@@ -502,7 +502,12 @@ The <dfn attribute for="XRProjectionLayer">ignoreDepthValues</dfn> attribute, if
 is <code>false</code> it indicates that the content of the depth buffer attachment will be used by the
 [=XR Compositor=] and is expected to be representative of the scene rendered into the layer.
 
-The <dfn attribute for="XRProjectionLayer">deltaPose</dfn> attribute describes the incremental application-applied transform, if any, since the previous frame that affects the view. If the user agent or device does not support this attribute or the "[=feature descriptor/space-warp=]" [=feature descriptor=] was not requested, they should return <code>null</code> on getting, and setting should be a <code>no-op</code>. When artificial locomotion happens, the experience transforms the whole world from one application pose to another pose between frames and {{XRProjectionLayer/deltaPose}} MUST be populated with the difference in the scene's world position. The {{XRProjectionLayer/deltaPose}} should be identity when there is no transformation between the previous and the current {{XRFrame}}.
+The <dfn attribute for="XRProjectionLayer">deltaPose</dfn> attribute describes the incremental application-applied transform, if any, since the previous frame that affects the view.
+If the user agent or device does not support this attribute or the "[=feature descriptor/space-warp=]" [=feature descriptor=] was not requested, the user agent should return <code>null</code> on getting, and setting should be a <code>no-op</code>.
+A <code>null</code> value indicates that no application-applied transform is being supplied.
+The value of {{XRProjectionLayer/deltaPose}} persists until the author changes it; the user agent does not reset it at the end of an {{XRSession/requestAnimationFrame()}} callback.
+When artificial locomotion transforms the world from one application pose to another between frames, setting {{XRProjectionLayer/deltaPose}} to the difference between those poses helps the [=XR Compositor=] synthesize missing frames.
+Authors should update {{XRProjectionLayer/deltaPose}} when that transform changes and set it to <code>null</code> or an identity transform for frames where no application-applied transform occurred.
 
 XRQuadLayer {#xrquadlayertype}
 -----------
@@ -775,8 +780,6 @@ The {{XRWebGLSubImage}} object is used during rendering of the layer.
   readonly attribute unsigned long colorTextureHeight;
   readonly attribute unsigned long? depthStencilTextureWidth;
   readonly attribute unsigned long? depthStencilTextureHeight;
-  readonly attribute unsigned long? motionVectorTextureWidth;
-  readonly attribute unsigned long? motionVectorTextureHeight;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


webxrlayers-1.bs
```diff
@@ -502,18 +502,7 @@ The <dfn attribute for="XRProjectionLayer">ignoreDepthValues</dfn> attribute, if
...
-locomotion happens, the experience transforms the whole world from one application pose to another pose between frames
-and {{XRProjectionLayer/deltaPose}} MUST be populated with the difference in the scene's world position.
-The {{XRProjectionLayer/deltaPose}} should be identity when there is no transformation between the previous and the current {{XRFrame}}.
-
-</div>
+The <dfn attribute for="XRProjectionLayer">deltaPose</dfn> attribute describes the incremental application-applied transform, if any, since the previous frame that affects the view. If the user agent or device does not support this attribute or the "[=feature descriptor/space-warp=]" [=feature descriptor=] was not requested, they should return <code>null</code> on getting, and setting should be a <code>no-op</code>. When artificial locomotion happens, the experience transforms the whole world from one application pose to another pose between frames and {{XRProjectionLayer/deltaPose}} MUST be populated with the difference in the scene's world position. The {{XRProjectionLayer/deltaPose}} should be identity when there is no transformation between the previous and the current {{XRFrame}}.
```
toji: > MUST be populated with the difference in the scene's world position.

This is describing a value that the developer is setting, not the browser, right? I believe we try to not use RFC statements like `MUST` for inputs from the developer. It's more to describe guarantees from the user agent. Would suggest rewording to indicate why filling this value out is a benefit to the developer.

toji: Also not clear if the developer is expected to reset this manually to an identity transform the frame after artificial locomotion occurs, or if the user agent resets it each frame?

webxrlayers-1.bs
```diff
@@ -796,30 +785,22 @@ The <dfn attribute for="XRWebGLSubImage">colorTexture</dfn> attribute returns th
...
-
 The <dfn attribute for="XRWebGLSubImage">motionVectorTexture</dfn> attribute returns the motion [=opaque texture=] for the {{XRProjectionLayer}}.
 If the {{XRSession}} was created without the [=feature descriptor/space-warp=] [=feature descriptor=] or the layer is not a {{XRProjectionLayer}}, this attribute MUST return null.
 
-</div>
-
 The <dfn attribute for="XRWebGLSubImage">imageIndex</dfn> attribute returns the offset into the texture array. Valid only for layers
 that were requested with {{texture-array}}.
 
 The <dfn attribute for="XRWebGLSubImage">colorTextureWidth</dfn> and <dfn attribute for="XRWebGLSubImage">depthStencilTextureHeight</dfn> attributes
```
toji: Mixed pairing of color width and depth height here

webxrlayers-1.bs
```diff
@@ -796,30 +785,22 @@ The <dfn attribute for="XRWebGLSubImage">colorTexture</dfn> attribute returns th
...
 return the width and height in pixels of the GL depth attachments
 
-<div class="unstable">
-
 The <dfn attribute for="XRWebGLSubImage">depthStencilTextureWidth</dfn> and <dfn attribute for="XRWebGLSubImage">colorTextureHeight</dfn> attributes
```
toji: Same here, but reversed

webxrlayers-1.bs
```diff
@@ -796,30 +785,22 @@ The <dfn attribute for="XRWebGLSubImage">colorTexture</dfn> attribute returns th
...
 return the width and height in pixels of the GL color attachments. If the layer was created without depth/stencil, this attribute returns null.
 
 The <dfn attribute for="XRWebGLSubImage">motionVectorTextureWidth</dfn> and <dfn attribute for="XRWebGLSubImage">motionVectorTextureHeight</dfn> attributes
 return the width and height in pixels of the GL motion vector attachments.
 If the {{XRSession}} was created without the [=feature descriptor/space-warp=] [=feature descriptor=] or the layer is not a {{XRProjectionLayer}}, these attributes MUST return null.
```
toji: Must return `null` for a numeric value? Seems like `0` or at the very least `undefined` is more appropriate?

webxrlayers-1.bs
```diff
@@ -2097,18 +2078,17 @@ Other layers types MUST continue to be sorted as before.
...
 
 To enable [=space warp=] the {{XRSession}} MUST be created with the "<dfn for="feature descriptor">space-warp</dfn>" [=feature descriptor=].
 If "[=space-warp=]" is enabled, the [=XR Compositor=] MUST make use of depth values, {{XRProjectionLayer/ignoreDepthValues}} MUST be set to `false` and {{XRWebGLBinding/usesDepthValues}} MUST be set to `true`.
 
-The {{XRWebGLSubImage/motionVectorTexture}} MUST be in {{RGBA16F}} format. The author SHOULD fill in the RGB components of this texture with the speed in meters per second of that area with the red pixel corresponding with the x axis, green the y axis and blue the z axis.
+When the [=feature descriptor/space-warp=] [=feature descriptor=] is enabled, the {{XRWebGLSubImage/depthStencilTexture}} for an {{XRProjectionLayer}} MUST be populated by the experience. The {{XRWebGLSubImage/depthStencilTexture}} and {{XRWebGLSubImage/motionVectorTexture}} will have the same dimensions, which may be different from the dimensions of the {{XRWebGLSubImage/colorTexture}}.
```
toji: If the motionVector and depthStencil textures are always the same size, do we still need to add a new set of width/height attribs?

toji (commented):
A few comments, sorry if some of this probably should have been raised earlier than this point.

webxrlayers-1.bs
cabanier: they are always the same.

cabanier: I removed them

webxrlayers-1.bs
cabanier: I clarified that the author needs to set/reset it every frame. This does feel a bit awkward..

webxrlayers-1.bs
cabanier: these were removed

webxrlayers-1.bs
```diff
@@ -2086,7 +2091,12 @@ the [=XR Compositor=] can do high quality frame extrapolation and reprojection w
...
+When the [=feature descriptor/space-warp=] [=feature descriptor=] is enabled, the experience should render representative depth values into the {{XRWebGLSubImage/depthStencilTexture}} for each {{XRProjectionLayer}} it submits.
+The {{XRWebGLSubImage/depthStencilTexture}} and {{XRWebGLSubImage/motionVectorTexture}} will have the same dimensions, which may be different from the dimensions of the {{XRWebGLSubImage/colorTexture}}.
+Authors can use {{XRWebGLSubImage/depthStencilTextureWidth}} and {{XRWebGLSubImage/depthStencilTextureHeight}} to determine the dimensions of both the depth/stencil and motion vector attachments.
+
+NOTE: Since the depth/stencil attachment can have different dimensions than the color attachment, it is not intended to be attached to the framebuffer used to render into the {{XRWebGLSubImage/colorTexture}}.
+Authors should render depth/stencil information separately at the depth/stencil attachment dimensions.
```
toji, line 2099: Hm... Does Meta return a depth texture with a different size than the color texture? I can see this being a subtle breaking issue for many devs.

toji (approved):
LGTM, thanks!

webxrlayers-1.bs
cabanier: In the case of space warp, yes. Otherwise no.

toji: I wonder if that's a reasonable thing to make a requirement in the spec? The depth/stencil resolution must match the color unless space warp is enabled, at which point it may (and probably should?) be different. At least then the risk of it breaking for developers between devices is reduced?

cabanier: The algorithm for allocating depth textures already states this. Do you think we need to make it more explicit when describing the depth texture dimensions?

toji: No, that's fine. As long as it's captured somewhere. Thanks!

